PR #2, opened 2020-01-24
MDEXP-1: Streaming API for working with CSV files in batches

mrdpopov:
# Purpose

Create an abstraction for working working with batches of Folio records. 

# Approach

Use java streams to split up the incoming UUIDs into batches of given size irrespective of where UUIDs come from. The default implementation assumes the UUIds come from a file saved to the local file system.

Basic unit test coverage has been provided.

# One more thing...

There is a utility class for generating large CSV files.


Igor-Gorchakov:
Looks good enough for PoC, we can reuse it

Igor-Gorchakov:
We will reuse it in slightly simplified way

src/main/java/org/folio/domain/reader/SourceStreamReader.java
```diff
@@ -0,0 +1,9 @@
+package org.folio.domain.reader;
+
+import java.io.IOException;
+import java.util.stream.Stream;
+
+public interface SourceStreamReader<T> {
+
+  Stream<T> getSourceStream() throws IOException;
```
Igor-Gorchakov: We will use FileStorageService instead of this abstraction

src/main/java/org/folio/service/DefaultRecordLoaderService.java
```diff
@@ -0,0 +1,56 @@
...
+
+  public DefaultRecordLoaderService(FileUploadService fileUploadService) {
+    this.fileUploadService = fileUploadService;
+  }
+
+  @Override
+  public Stream<Batch<? extends FolioRecord>> getBatchRecordsStream(int batchSize) throws IOException {
+    Stream<List<String>> sourceStream = fileUploadService.getSourceStream(batchSize);
+    return sourceStream
+      .map(this::callSrs)
```
Igor-Gorchakov: As we work in the async world the method callSrs returns Future. Here we can not rely on async call in sync method.

PR #3, opened 2020-01-27
MDEXP-35 Create the main export flow

Igor-Gorchakov:


NEWS.md
```diff
@@ -1 +1,3 @@
-* Incremental Release Notes here.
+## 2020-01-15 v0.0.1-SNAPSHOT
```
KVupp: Let's have only release notes here. Which will be updated during the release

pom.xml
```diff
@@ -3,7 +3,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>org.folio</groupId>
   <artifactId>mod-data-export</artifactId>
-  <version>1.0.0-SNAPSHOT</version>
+  <version>0.0.1-SNAPSHOT</version>
```
KVupp: I think this has to be 1.0.0-SNAPSHOT. while releasing we set it to 1.0.0

ramls/dataExport.raml
```diff
@@ -0,0 +1,53 @@
...
+  errors: !include raml-util/schemas/errors.schema
+  exportRequest: !include schemas/exportRequest.json
+
+traits:
+  validate: !include raml-util/traits/validation.raml
+  language: !include raml-util/traits/language.raml
+  pageable:  !include raml-util/traits/pageable.raml
+  searchable: !include raml-util/traits/searchable.raml
+
+resourceTypes:
```
KVupp: let's add only those are needed

src/test/java/org/folio/rest/impl/AbstractRestTest.java
```diff
@@ -0,0 +1,107 @@
...
+import org.junit.BeforeClass;
+import org.junit.Rule;
+
+import java.io.IOException;
+
+import static org.folio.rest.RestVerticle.OKAPI_HEADER_TENANT;
+
+public abstract class AbstractRestTest {
+
+  protected static final String OKAPI_TENANT_HEADER = "x-okapi-tenant";
```
KVupp: RMB already has all  these variables defined. you can reuse them instead of re-defining
import static org.folio.rest.RestVerticle.OKAPI_HEADER_TENANT; etc..

src/main/java/org/folio/util/OkapiConnectionParams.java
```diff
@@ -0,0 +1,36 @@
+package org.folio.util;
+
+import java.util.Map;
+
+public final class OkapiConnectionParams {
```
KVupp: I think this class is redundant. okapiheaders is already a map and can  be used only to make further calls  or fetch a httpClient when needed

src/test/java/org/folio/rest/impl/AbstractRestTest.java
```diff
@@ -0,0 +1,107 @@
...
+  protected static final String TOKEN = "token";
+  private static final String HTTP_PORT = "http.port";
+  private static final String HOST = "http://localhost:";
+
+  protected static RequestSpecification requestSpecification;
+  private static int port;
+  private static Vertx vertx;
+
+  @Rule
+  public WireMockRule mockServer = new WireMockRule(
```
KVupp: we can create a mock server and define mock routes using vertx. What is the advantage of using Wiremock?

Igor-Gorchakov: +

src/main/java/org/folio/util/OkapiConnectionParams.java
Igor-Gorchakov: Let's stay with this class. It's always easer to extract a header using the predefined method instead of working with HashMap

src/test/java/org/folio/rest/impl/AbstractRestTest.java
Igor-Gorchakov: done

ramls/dataExport.raml
Igor-Gorchakov: done

pom.xml
Igor-Gorchakov: done

NEWS.md
Igor-Gorchakov: done

src/test/java/org/folio/rest/impl/AbstractRestTest.java
Igor-Gorchakov: done

Igor-Gorchakov:
I will squash&merge this PR with the only one commit, to do not pollute git history

PR #4, opened 2020-02-03
MDEXP-6 Create Inventory Instance to MARC Bib record mapping rules

KVupp:
## PURPOSE
https://issues.folio.org/browse/MDEXP-6

## APPROACH
- Mapped files using the file https://docs.google.com/spreadsheets/d/11lGBiPoetHuC3u-onVVLN4Mj5KtVHqJaQe4RqCxgGzo/edit#gid=1166940623
- Left a few more questions on the ticket to be mapped

##TODO
- Indicators handlings is yet to be decided
- There is also an case handling required in indicators, need to decide on a approach on that

Note: I also added PR template, it will be helpful if we add more details in each PR. Please try to fill the PR template whenever you are creating a PR


Igor-Gorchakov:
Please, format json file, use https://jsonformatter.curiousconcept.com/

PR #5, opened 2020-02-06
MDEXP-43-44-45 File upload functionality

Igor-Gorchakov:
## Purpose
https://issues.folio.org/browse/MDEXP-43
https://issues.folio.org/browse/MDEXP-44
https://issues.folio.org/browse/MDEXP-45

## APPROACH
Use RMB streaming capabilities to upload files. Use relational database columns.


src/main/resources/templates/db_scripts/schema.json
```diff
@@ -0,0 +1,2 @@
+{
```
Igor-Gorchakov: Keep schema empty to let RMB to create schema for a tenant. 
If no schema.json provided, then RMB does not create a tenant-specific schema.

Igor-Gorchakov:
I will fix tests later

KVupp:
I have few concerns on this PR Igor
1) Not using jsonb vs columns, we don't know if it is approved at TC , we may have to check if it is folio wide accepted before using it
2) Just using the Insert/update statements is something that will become difficult to maintain. 

To summarize, we need to check if this will be acceptable at FOLIO level, until that we may have to take the old route to keep jsonb

mrdpopov:
LGTM

Igor-Gorchakov:
> I have few concerns on this PR Igor
> 
> 1. Not using jsonb vs columns, we don't know if it is approved at TC , we may have to check if it is folio wide accepted before using it
> 2. Just using the Insert/update statements is something that will become difficult to maintain.
> 
> To summarize, we need to check if this will be acceptable at FOLIO level, until that we may have to take the old route to keep jsonb

Finally, we will work with 2 tables, table for FileDefinitions and a table for Jobs. I don't think it makes a big effort to manage such lightweight relationship tables writing queries manually, pub-sub is a good example.

descriptors/ModuleDescriptor-template.json
```diff
@@ -17,6 +17,39 @@
...
+          "modulePermissions": [
+          ]
+        },
+        {
+          "methods": [
+            "POST"
+          ],
+          "pathPattern": "/data-export/fileDefinitions/{fileDefinitionId}",
+          "permissionsRequired": [
+            "data-export.filedefinitions.uploadfile"
```
KVupp: please name the permissions in line with othe rmodules. EAch entry post is usually named item.post

ramls/dataExport.raml
```diff
@@ -44,3 +44,79 @@ traits:
...
+            description: "Not found"
+            body:
+              text/plain:
+                example: "Not found"
+          500:
+            description: "Internal server error"
+            body:
+              text/plain:
+                example: "Internal server error"
+      post:
```
KVupp: Isn't it better to use a separate endpoint for upload? It's a little deceiving that GET  /filedefinitions/{fileID} will get file definition and POST will do an upload of a file
we can do POST /filedefinitions/{fileID}/upload for file uploads

ramls/dataExport.raml
```diff
@@ -44,3 +44,79 @@ traits:
           body:
             text/plain:
               example: "Internal server error"
+  /fileDefinitions:
+    displayName: File upload service
+    description: Method to create file definition
```
KVupp: Description can be expanded a little

ramls/schemas/fileDefinition.json
```diff
@@ -6,15 +6,43 @@
...
+    "jobExecutionId": {
+      "description": "Job execution id",
+      "type": "string",
+      "ref": "../common/uuid.json"
+    },
+    "sourcePath": {
       "description": "Path to the file",
       "type": "string"
+    },
+    "createdDate": {
```
KVupp: can we use the metadata schema like in other modules? We may have to display who updated it and created it etc..

KVupp: This will also be another difference from other modules, we'll not have access to such data, of who updated it when

src/main/java/org/folio/dao/util/ConnectionProvider.java
```diff
@@ -0,0 +1,62 @@
...
+
+  /**
+   * Returns database connection for the given tenant
+   *
+   * @param vertx  vertx instance
+   * @param tenant given tenant to which schema connection should be provided
+   * @return Connection to database
+   * @throws SQLException if error occurs while trying to obtain access to database
+   */
+  public static Connection getConnection(Vertx vertx, String tenant) throws SQLException {
```
KVupp: we are not using a connection pool. That will lead to a lot an expensive operation of fetching connection each time

src/main/java/org/folio/dao/impl/FileDefinitionDaoImpl.java
```diff
@@ -0,0 +1,107 @@
...
+      String query = format(INSERT_QUERY, convertToPsqlStandard(tenantId), TABLE);
+      JsonArray params = new JsonArray()
+        .add(fileDefinition.getId())
+        .add(fileDefinition.getFileName())
+        .add(fileDefinition.getJobExecutionId())
+        .add(fileDefinition.getSourcePath())
+        .add(Timestamp.from(fileDefinition.getCreatedDate().toInstant()).toString())
+        .add(fileDefinition.getStatus())
+        .add(fileDefinition.getSize());
+      pgClientFactory.getInstance(tenantId).execute(query, params, promise);
```
KVupp: The DB connection is not being closed here

src/main/java/org/folio/dao/impl/FileDefinitionDaoImpl.java
```diff
@@ -0,0 +1,107 @@
...
+      String query = format(UPDATE_BY_ID_QUERY, convertToPsqlStandard(tenantId), TABLE);
+      JsonArray params = new JsonArray()
+        .add(fileDefinition.getFileName())
+        .add(fileDefinition.getJobExecutionId())
+        .add(fileDefinition.getSourcePath())
+        .add(Timestamp.from(fileDefinition.getCreatedDate().toInstant()).toString())
+        .add(fileDefinition.getStatus())
+        .add(fileDefinition.getSize())
+        .add(fileDefinition.getId());
+      pgClientFactory.getInstance(tenantId).execute(query, params, promise);
```
KVupp: same here, connection not being closed

src/main/java/org/folio/dao/util/ConnectionProvider.java
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  private static Connection getConnectionInternal(JsonObject connectionConfig) throws SQLException {
+    String username = connectionConfig.getString(CONFIG_USERNAME_KEY);
+    String password = connectionConfig.getString(CONFIG_PASS_KEY);
+    String host = connectionConfig.getString(CONFIG_HOST_KEY);
+    String port = String.valueOf(connectionConfig.getInteger(CONFIG_PORT_KEY));
+    String database = connectionConfig.getString(CONFIG_DATABASE_KEY);
+    String connectionUrl = format("%s://%s:%s/%s", JDBC_DRIVER, host, port, database);
+    return DriverManager.getConnection(connectionUrl, username, password);
```
KVupp: Instead of implementing our our Client, I think we can use the RMB postgres client with the execute method?

descriptors/ModuleDescriptor-template.json
```diff
@@ -17,6 +17,39 @@
...
+          "modulePermissions": [
+          ]
+        },
+        {
+          "methods": [
+            "POST"
+          ],
+          "pathPattern": "/data-export/fileDefinitions",
+          "permissionsRequired": [
+            "data-export.filedefinitions.post"
```
KVupp: can we call it item.post

descriptors/ModuleDescriptor-template.json
```diff
@@ -17,6 +17,39 @@
...
+          ],
+          "modulePermissions": [
+          ]
+        },
+        {
+          "methods": [
+            "POST"
+          ],
+          "pathPattern": "/data-export/fileDefinitions",
+          "permissionsRequired": [
```
KVupp: Also, it would be good to create a permission set containing all interface permissions, so that UI can use that

Igor-Gorchakov:
Agreed to close PR, **do not remove the branch**. Related PR is https://github.com/folio-org/mod-data-export/pull/6

PR #6, opened 2020-02-13
Mdexp 43 44 45 File upload funtionality

Igor-Gorchakov:
## Purpose
https://issues.folio.org/browse/MDEXP-43
https://issues.folio.org/browse/MDEXP-44
https://issues.folio.org/browse/MDEXP-45

## APPROACH
Use RMB streaming capabilities to upload files. Use approach with regular 'jsonb' column.


ramls/dataExport.raml
```diff
@@ -11,15 +11,15 @@ documentation:
...
   exportRequest: !include schemas/exportRequest.json
 
 traits:
   validate: !include raml-util/traits/validation.raml
-  language: !include raml-util/traits/language.raml
 
 /data-export:
   /export:
-    displayName: Data export service
+    displayName: Export manager service
```
KVupp: I think this was the description Magda was mentioning last time that we need to be more elaborate. Can we confirm with her once

ramls/dataExport.raml
```diff
@@ -44,3 +44,80 @@ traits:
...
+          description: "Unprocessable Entity"
+          body:
+            application/json:
+              type: errors
+        500:
+          description: "Internal server error"
+          body:
+            text/plain:
+              example: "Internal server error"
+    /{fileDefinitionId}:
```
KVupp: It would be nice to add a sample file.

src/main/java/org/folio/dao/impl/FileDefinitionDaoImpl.java
```diff
@@ -0,0 +1,72 @@
...
+  public FileDefinitionDaoImpl(@Autowired PostgresClientFactory pgClientFactory) {
+    this.pgClientFactory = pgClientFactory;
+  }
+
+  @Override
+  public Future<Optional<FileDefinition>> getById(String id, String tenantId) {
+    Promise<Results<FileDefinition>> promise = Promise.promise();
+    try {
+      Criteria idCrit = constructCriteria(ID_FIELD, id);
+      pgClientFactory.getInstance(tenantId).get(TABLE, FileDefinition.class, new Criterion(idCrit), false, promise);
```
KVupp: can we use PgUtil? All this exception handling is already done there, all this ocde will be reduced to a single line:
```suggestion
       PgUtil.getById(TABLE, FileDefinition.class, id, okapiHeaders, vertxContext, GetDataExportFileDefinitionsByFileDefinitionIdResponse.class, asyncResultHandler);
```

KVupp: You need not return an Optional in that case, the method will directly throw a 404

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -0,0 +1,16 @@
+{
+  "tables": [
+    {
+      "tableName": "file_definitions",
+      "fromModuleVersion": "mod-data-export-1.0.0",
+      "withMetadata": true,
+      "withAuditing": false,
+      "uniqueIndex": [
```
KVupp: UUIDs are unique universally, this is not required

src/main/java/org/folio/dao/impl/FileDefinitionDaoImpl.java
```diff
@@ -0,0 +1,72 @@
...
+    this.pgClientFactory = pgClientFactory;
+  }
+
+  @Override
+  public Future<Optional<FileDefinition>> getById(String id, String tenantId) {
+    Promise<Results<FileDefinition>> promise = Promise.promise();
+    try {
+      Criteria idCrit = constructCriteria(ID_FIELD, id);
+       PgUtil.getById(TABLE, FileDefinition.class, id, okapiHeaders, vertxContext, GetDataExportFileDefinitionsByFileDefinitionIdResponse.class, asyncResultHandler);
+    } catch (Exception e) {
```
KVupp: all the try catch is also not required. the Criteria is not required. Pgutil takes care of it. 
Example:
https://github.com/folio-org/mod-finance-storage/blob/master/src/main/java/org/folio/rest/impl/BudgetAPI.java#L36

KVupp: same for the other places please. where save and update are being used

src/main/java/org/folio/dao/impl/FileDefinitionDaoImpl.java
Igor-Gorchakov: Let's do not use PgUtil at all. PgUtil does not useful enough to work with the database, because it is API is tightly coupled to the Response and forces DAO to form responses:)   Data access layer should be responsible only for data retrieval.

Igor-Gorchakov:
> Except using the PgUtil I have nothing else that is major.
> In interest of time , we can improve few minor things later

We do not use PgUtil. The API of this util is tightly coupled to the _Class<? extends ResponseDelegate> response_, so it is not useful in DAOs at all. This approach breaks an concept of DAOs, the data access layer should be responsible only for data selection/insertion but not building the HTTP Response :) To build a Response we use rest.impl resource.

KVupp:
> > Except using the PgUtil I have nothing else that is major.
> > In interest of time , we can improve few minor things later
> 
> We do not use PgUtil. The API of this util is tightly coupled to the _Class<? extends ResponseDelegate> response_, so it is not useful in DAOs at all. This approach breaks an concept of DAOs, the data access layer should be responsible only for data selection/insertion but not building the HTTP Response :) To build a Response we use rest.impl resource.

ok let's continue with the DAO pattern for now

PR #8, opened 2020-02-13
MDEXP-36: RecordLoaderService: retrieve MARC records from SRS

mrdpopov:
## Purpose

For a batch of UUID retrieve MARC records by calling SRS

## Approach

Synchronous implementation of RecordLoaderService that accepts list of uuids and returns the corresponding marc records if they exist in SRS.

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -1,7 +1,12 @@
 {
   "id": "${artifactId}-${version}",
   "name": "Data Export Module",
-  "requires": [],
+  "requires": [
+    {
+      "id": "source-storage-records",
+      "version": "1.0"
+    }
+  ],
   "provides": [
     {
       "id": "data-export",
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -16,6 +16,7 @@
             "data-export.export.post"
           ],
           "modulePermissions": [
+            "source-storage.records.get"
           ]
         },
         {
--- a/src/main/java/org/folio/clients/OkapiClients.java
+++ b/src/main/java/org/folio/clients/OkapiClients.java
@@ -24,6 +24,4 @@ public class OkapiClients {
     }
     return sourceRecordStorageClient;
   }
-
-
 }
--- a/src/main/java/org/folio/clients/OkapiClients.java
+++ b/src/main/java/org/folio/clients/OkapiClients.java
@@ -1,0 +1,29 @@
+package org.folio.clients;
+
+import org.folio.clients.impl.SourceRecordStorageClient;
+import org.folio.util.OkapiConnectionParams;
+
+public class OkapiClients {
+
+  private final OkapiConnectionParams okapiConnectionParams;
+
+  private SynchronousOkapiClient sourceRecordStorageClient;
+
+
+  private OkapiClients(OkapiConnectionParams okapiConnectionParams) {
+    this.okapiConnectionParams = okapiConnectionParams;
+  }
+
+  public static OkapiClients create(OkapiConnectionParams okapiConnectionParams) {
+    return new OkapiClients(okapiConnectionParams);
+  }
+
+  public SynchronousOkapiClient getSourceRecordStorageClient() {
+    if (sourceRecordStorageClient == null) {
+      sourceRecordStorageClient = new SourceRecordStorageClient(okapiConnectionParams);
+    }
+    return sourceRecordStorageClient;
+  }
+
+
+}
--- a/src/main/java/org/folio/clients/SynchronousOkapiClient.java
+++ b/src/main/java/org/folio/clients/SynchronousOkapiClient.java
@@ -1,17 +1,6 @@
 package org.folio.clients;
 
-import static org.folio.rest.RestVerticle.OKAPI_HEADER_TENANT;
-import static org.folio.rest.RestVerticle.OKAPI_HEADER_TOKEN;
-
-import java.io.IOException;
-import java.lang.invoke.MethodHandles;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.function.Function;
-
-import javax.ws.rs.core.MediaType;
-
+import io.vertx.core.json.JsonObject;
 import org.apache.http.HttpHeaders;
 import org.apache.http.client.methods.CloseableHttpResponse;
 import org.apache.http.client.methods.HttpGet;
@@ -22,7 +11,15 @@ import org.folio.util.OkapiConnectionParams;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import io.vertx.core.json.JsonObject;
+import javax.ws.rs.core.MediaType;
+import java.io.IOException;
+import java.lang.invoke.MethodHandles;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+import static org.folio.rest.RestVerticle.OKAPI_HEADER_TENANT;
+import static org.folio.rest.RestVerticle.OKAPI_HEADER_TOKEN;
 
 /**
  * Base class for okapi clients that use blocking method for http calls.
@@ -31,9 +28,11 @@ public abstract class SynchronousOkapiClient {
 
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
   protected final OkapiConnectionParams okapiConnectionParams;
+  private final CloseableHttpClient httpClient;
 
   public SynchronousOkapiClient(OkapiConnectionParams okapiConnectionParams) {
     this.okapiConnectionParams = okapiConnectionParams;
+    httpClient = HttpClients.createDefault();
   }
 
   public Optional<JsonObject> getByIds(List<String> ids) {
@@ -41,8 +40,7 @@ public abstract class SynchronousOkapiClient {
     setCommonHeaders(httpGet);
     prepareRequest(httpGet, ids);
 
-    try (CloseableHttpClient httpClient = HttpClients.createDefault();
-         CloseableHttpResponse response = httpClient.execute(httpGet)) {
+    try (CloseableHttpResponse response = httpClient.execute(httpGet)) {
       return Optional.ofNullable(postProcess(response));
     } catch (IOException e) {
       log.error("Exception while calling " + httpGet.getURI(), e);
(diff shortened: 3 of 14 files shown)


src/main/java/org/folio/clients/impl/SourceRecordStorageClient.java
```diff
@@ -0,0 +1,56 @@
...
+ * Retrieves collection of underlying SRS records as a source of truth.
+ * Retrieves collection of Inventory records that do not have underlying SRS records
+ *
+ * @link https://github.com/folio-org/mod-source-record-storage
+ */
+@Component
+public class SourceRecordStorageClient extends SynchronousOkapiClient {
+
+  private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+  public static final String GET_RECORD_PATTERN = "%s/source-storage/formattedRecords/%s?identifier=INSTANCE";
```
KVupp: This permission needs to be added to Moduledescriptor

mrdpopov:
TODO:

** Todo:

* Use the following  URI:
/source-storage/records?query=(externalIdsHolder.instanceId==1f245392-2468-4c41-b6a3-164edeee1349 or externalIdsHolder.instanceId==947b5847-a3ec-44ad-9dc4-c8b8168a8d0b)

src/main/java/org/folio/clients/impl/SourceRecordStorageClient.java
```diff
@@ -0,0 +1,56 @@
...
+
+  public SourceRecordStorageClient(OkapiConnectionParams okapiConnectionParams) {
+    super(okapiConnectionParams);
+  }
+
+  @Override
+  protected void prepareRequest(HttpRequestBase requestBase, String id) {
+    requestBase.setURI(URI.create(String.format(GET_RECORD_PATTERN,
```
KVupp: Also like discussed on slack already, try if we can use API similar to 
`curl --location --request GET 'https://folio-testing-okapi.aws.indexdata.com:443/inventory/instances?query=id=%22e6bc03c6-c137-4221-b679-a7c5c31f986c%22%20or%20id=%225bf370e0-8cca-4d9c-82e4-5170ab2a0a39%22' \
--header 'x-okapi-token: {{okapitoken}}'`
It will help in reducing total number of calls

KVupp: Last time I checked you can use a maximum of 15 ids per request with or

src/main/java/org/folio/clients/SynchronousOkapiClient.java
```diff
@@ -0,0 +1,63 @@
...
+
+  private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+  protected final OkapiConnectionParams okapiConnectionParams;
+
+  public SynchronousOkapiClient(OkapiConnectionParams okapiConnectionParams) {
+    this.okapiConnectionParams = okapiConnectionParams;
+  }
+
+  public Optional<JsonObject> getById(String id) {
+    HttpGet httpGet = new HttpGet();
```
KVupp: here we are establishing a new HTTP Connection each time we make a get request. I think we can use one connection, and close it after all the calls are made

src/main/java/org/folio/clients/impl/SourceRecordStorageClient.java
mrdpopov: That can been added not, my initial intention was to do it later.

mrdpopov: Done.

src/main/java/org/folio/clients/impl/SourceRecordStorageClient.java
mrdpopov: The code has been chaged to use the URI stated above.

src/main/java/org/folio/clients/SynchronousOkapiClient.java
mrdpopov: Changed.

descriptors/ModuleDescriptor-template.json
```diff
@@ -16,6 +16,7 @@
             "data-export.export.post"
           ],
           "modulePermissions": [
+            "source-storage.records.get"
```
KVupp: We need the interface to be added as a dependency too

KVupp: source-storage-records 1.0

src/main/java/org/folio/clients/SynchronousOkapiClient.java
```diff
@@ -0,0 +1,63 @@
...
+    setCommonHeaders(httpGet);
+    prepareRequest(httpGet, ids);
+
+    try (CloseableHttpResponse response = httpClient.execute(httpGet)) {
+      return Optional.ofNullable(postProcess(response));
+    } catch (IOException e) {
+      log.error("Exception while calling " + httpGet.getURI(), e);
+      return Optional.empty();
+    }
```
KVupp: should be closing the response , it is leaking connections
check: https://stackoverflow.com/questions/19920037/why-does-me-use-httpclients-createdefault-as-httpclient-singleton-instance-exe

mrdpopov: The response is being closed, it's in try with resources block.

mrdpopov: If there is a need to close the client, I think, it should be handled elsewhere.

KVupp: We definitely need to close the client.  What is elsewhere?

mrdpopov: Closing it In the calling code would be my suggestion.

KVupp: ok I went through the code again, seems fair in this situation

KVupp: Lets make to check this when we write that code

src/main/java/org/folio/clients/OkapiClients.java
```diff
@@ -0,0 +1,27 @@
+package org.folio.clients;
+
+import org.folio.clients.impl.SourceRecordStorageClient;
+import org.folio.util.OkapiConnectionParams;
+
+public class OkapiClients {
```
Igor-Gorchakov: Please rename it to OkapiClientFactory

src/main/java/org/folio/clients/SynchronousOkapiClient.java
```diff
@@ -0,0 +1,63 @@
...
+
+import static org.folio.rest.RestVerticle.OKAPI_HEADER_TENANT;
+import static org.folio.rest.RestVerticle.OKAPI_HEADER_TOKEN;
+
+/**
+ * Base class for okapi clients that use blocking method for http calls.
+ */
+public abstract class SynchronousOkapiClient {
+
+  private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
```
Igor-Gorchakov: Please, instantiate logger in our usual way : LoggerFactory.getLogger(SynchronousOkapiClient.class);

src/test/java/org/folio/rest/impl/HttpServerTestBase.java
```diff
@@ -0,0 +1,66 @@
...
+      .addHeader(OKAPI_HEADER_TENANT, TENANT_ID)
+      .setBaseUri(OKAPI_URL)
+      .addHeader("Accept", "text/plain, application/json")
+      .build();
+  }
+
+  @NotNull
+  protected OkapiConnectionParams getOkapiConnectionParams() {
+    Map<String, String> okapiHeaders = new HashMap<>();
+    okapiHeaders.put("x-okapi-url", "http://localhost:" + httpServer.actualPort());
```
Igor-Gorchakov: Use final String HOST property

src/main/java/org/folio/clients/OkapiClients.java
```diff
@@ -0,0 +1,27 @@
...
+  private OkapiClients(OkapiConnectionParams okapiConnectionParams) {
+    this.okapiConnectionParams = okapiConnectionParams;
+  }
+
+  public static OkapiClients create(OkapiConnectionParams okapiConnectionParams) {
+    return new OkapiClients(okapiConnectionParams);
+  }
+
+  public SynchronousOkapiClient getSourceRecordStorageClient() {
+    if (sourceRecordStorageClient == null) {
```
Igor-Gorchakov: This initialization is safe only in the single-thread application. We probably will use multiple worker-threads, sourceRecordStorageClient should be initialized safety

src/main/java/org/folio/clients/impl/SourceRecordStorageClient.java
```diff
@@ -0,0 +1,74 @@
...
+ * <p>
+ * Retrieves collection of underlying SRS records as a source of truth.
+ * Retrieves collection of Inventory records that do not have underlying SRS records
+ *
+ * @link https://github.com/folio-org/mod-source-record-storage
+ */
+@Component
+public class SourceRecordStorageClient extends SynchronousOkapiClient {
+
+  private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
```
Igor-Gorchakov: LoggerFactory.getLogger(SourceRecordStorageClient.class);

src/main/java/org/folio/service/loader/BlockingRecordLoaderService.java
```diff
@@ -0,0 +1,55 @@
...
+    optionalRecords.ifPresent(records -> populateLoadResult(uuids, records, srsLoadResult));
+    return srsLoadResult;
+  }
+
+  private void populateLoadResult(List<String> queriedUuids, JsonObject foundRecords, SrsLoadResult loadResult) {
+    JsonArray records = foundRecords.getJsonArray("records");
+    Set<String> marcRecords = new HashSet<>();
+    Set<String> setSingleInstanceIdentifiers = new HashSet<>(queriedUuids);
+    for (Object o : records) {
+      if (o instanceof JsonObject) {
```
Igor-Gorchakov: If a record is not an instance of JsonObject, please Log it

src/main/java/org/folio/clients/SynchronousOkapiClient.java
mrdpopov: To me that was the usual way

src/main/java/org/folio/clients/impl/SourceRecordStorageClient.java
mrdpopov: Why is this better?

src/test/java/org/folio/rest/impl/HttpServerTestBase.java
mrdpopov: Done

PR #9, opened 2020-02-17
MDEXP-54 ExportManager - Implement Export flow

Igor-Gorchakov:
## Purpose

To have a basic export implementation for further enhancements

## Approach
Implemented ExportManager.exportBlocking method
Load Srs and Inventory records in partitions
MappingService, ExportService, LoaderService are stubbed
Covered with tests


src/main/java/org/folio/service/manager/exportmanager/ExportManagerImpl.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+  }
+
+  /**
+   * Runs the main export flow in blocking manner.
+   *
+   * @param identifiers list of inventory identifiers
+   */
+  protected void exportBlocking(List<String> identifiers) {
+    Lists.partition(identifiers, EXPORT_PARTITION_SIZE).forEach(partitionIdentifiers -> {
```
Igor-Gorchakov: TODO: Discuss algorithm

Igor-Gorchakov:
This is the latest version, please review @KVupp @mrdpopov

mrdpopov:
Please integrate it with the master branch.

src/main/java/org/folio/service/loader/MarcLoadResult.java
```diff
@@ -0,0 +1,31 @@
+package org.folio.service.loader;
+
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class MarcLoadResult {
+  private List<String> srsMarcRecords;
+  private List<String> singleInstanceIdentifiers;
```
KVupp: what does single signify? can we name this simply instanceIdentifiers.

src/main/java/org/folio/service/loader/MarcLoadResult.java
```diff
@@ -0,0 +1,31 @@
+package org.folio.service.loader;
+
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class MarcLoadResult {
```
KVupp: shouldn't this be a schema file? and be generated rather than create a POJO ourselves?

src/main/java/org/folio/service/upload/FileUploadService.java
```diff
@@ -40,8 +40,8 @@
   /**
    * Completes uploading for the given {@link FileDefinition}
    *
-   * @param fileDefinition  {@link FileDefinition}
-   * @param tenantId tenant id
+   * @param fileDefinition {@link FileDefinition}
```
KVupp: Not part of this PR, but why is the FileUploadService creating a FileDefinition?
Shouldn't it be part of FileDefinitionService?

src/main/java/org/folio/service/loader/BlockingRecordLoaderService.java
```diff
@@ -1,51 +1,56 @@
...
       JsonObject externalIdsHolder = record.getJsonObject("externalIdsHolder");
       if (externalIdsHolder != null) {
         String instanceId = externalIdsHolder.getString("instanceId");
-        setSingleInstanceIdentifiers.remove(instanceId);
+        singleInstanceIdentifiersSet.remove(instanceId);
       }
     }
     loadResult.setUnderlyingMarcRecords(marcRecords);
-    loadResult.setSingleInstanceIdentifiers(setSingleInstanceIdentifiers);
+    loadResult.setInstanceIdsWithoutSrs(new ArrayList<>(singleInstanceIdentifiersSet));
```
mrdpopov: Can we use arrayList instead of set to avoid that wrapping? (I know that set remained from previous implementation)

src/main/java/org/folio/service/loader/BlockingRecordLoaderService.java
```diff
@@ -1,51 +1,56 @@
...
+  @Autowired
+  private SourceRecordStorageClient client;
 
   @Override
-  public SrsLoadResult loadMarcRecords(List<String> uuids) {
-    Optional<JsonObject> optionalRecords = client.getByIds(uuids);
+  public SrsLoadResult loadMarcRecords(List<String> uuids, OkapiConnectionParams okapiConnectionParams) {
+    Optional<JsonObject> optionalRecords = client.getByIds(uuids, okapiConnectionParams);
     SrsLoadResult srsLoadResult = new SrsLoadResult();
     optionalRecords.ifPresent(records -> populateLoadResult(uuids, records, srsLoadResult));
```
KVupp: if there are no SRS records for any of the  UUIDs, the srsLoadResult will be entirely empty. The instanceswithoutsrs will not be populated. That should be done here

PR #11, opened 2020-02-25
MDEXP-61: Create JobExecution entity

IlliaDaliek:
## Purpose
Add new entity for jobExecution managment.

### DIFF
--- a/ramls/schemas/jobExecution.json
+++ b/ramls/schemas/jobExecution.json
@@ -29,7 +29,7 @@
       "description": "Number of exported records",
       "type": "integer"
     },
-    "endedRunning": {
+    "completedDate": {
       "description": "Date and time when the job execution completed",
       "type": "string",
       "format": "date-time"
@@ -48,11 +48,6 @@
         }
       }
     },
-    "metadata": {
-      "description": "Metadata",
-      "type": "object",
-      "$ref": "../../ramls/raml-util/schemas/metadata.schema"
-    },
     "status": {
       "description": "Status of files export",
       "type": "string",
@@ -64,9 +59,10 @@
         "FAIL"
       ]
     },
-    "errorCause": {
-      "description": "Describes the cause of execution errors",
-      "type": "string"
+    "metadata": {
+      "description": "Metadata",
+      "type": "object",
+      "$ref": "../../ramls/raml-util/schemas/metadata.schema"
     }
   },
   "required": [
--- a/ramls/schemas/jobExecution.json
+++ b/ramls/schemas/jobExecution.json
@@ -12,7 +12,7 @@
       "type": "string"
     },
     "exportedFiles": {
-      "description": "List of files that should be exported",
+      "description": "List of exported files",
       "type": "array",
       "uniqueItems": true,
       "items": {
@@ -20,7 +20,7 @@
         "$ref": "exportedFile.json"
       }
     },
-    "jobProfile": {
+    "jobProfileInfo": {
       "description": "Related JobProfile information",
       "type": "object",
       "$ref": "jobProfile.json"
--- a/ramls/schemas/jobExecution.json
+++ b/ramls/schemas/jobExecution.json
@@ -54,7 +54,7 @@
       ]
     },
     "errorCause":{
-      "description":"Status that describe error state of job execution",
+      "description":"Status that describes error state of job execution",
       "type":"string"
     }
   }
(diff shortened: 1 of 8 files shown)


ramls/schemas/jobExecution.json
```diff
@@ -0,0 +1,61 @@
...
+    "recordsCount":{
+      "description":"Number of exported records",
+      "type":"integer"
+    },
+    "endedRunning":{
+      "description":"Date and time when the job execution completed",
+      "type":"string",
+      "format":"date-time"
+    },
+    "runBy":{
```
KVupp: I think this information can be gathered from metadata , just include 
```suggestion
    "metadata": {
      "type": "object",
      "$ref": "../../../raml-util/schemas/metadata.schema",
      "readonly": true
    }
```

ramls/schemas/jobExecution.json
```diff
@@ -0,0 +1,61 @@
...
+          "type":"string"
+        },
+        "lastName":{
+          "description":"Last name",
+          "type":"string"
+        }
+      }
+    },
+    "status":{
+      "description":"Status of files upload",
```
KVupp: Should this be status of upload or export?

ramls/schemas/jobExecution.json
```diff
@@ -0,0 +1,61 @@
+{
+  "$schema":"http://json-schema.org/draft-04/schema#",
+  "description":"Job Execution Schema",
+  "type":"object",
+  "properties":{
+    "id":{
+      "description":"Unique identifier",
+      "$ref":"../../ramls/raml-util/schemas/uuid.schema"
+    },
+    "hrId":{
```
KVupp: what is the HR ID in this case? How do we plan on populating it?

ramls/schemas/jobExecution.json
```diff
@@ -0,0 +1,61 @@
...
+      "description":"Human readable id",
+      "type":"string"
+    },
+    "fileName":{
```
KVupp: Going forward we will be having a list of files, probably makes sense to make this an array

ramls/schemas/jobExecution.json
```diff
@@ -0,0 +1,61 @@
...
+      }
+    },
+    "status":{
+      "description":"Status of files upload",
+      "type":"string",
+      "enum":[
+        "NEW",
+        "IN_PROGRESS",
+        "SUCCESS",
+        "FAIL"
```
KVupp: We should also make status required field, and also add a default of "NEW"

ramls/schemas/jobExecution.json
Igor-Gorchakov: Please fix description, this should be about export

ramls/schemas/jobExecution.json
Igor-Gorchakov: Let's make it array

Igor-Gorchakov:
There are new updates in mockups, please update schema
https://drive.google.com/drive/folders/1da7RedCp1tRVdHw4QGNnu9B0q8mQSGyA

ramls/schemas/jobExecution.json
KVupp: add file Id and filename
```suggestion
    "file"{
    "fileId": 
    "fileName":
    }
```

ramls/schemas/jobExecution.json
```diff
@@ -0,0 +1,61 @@
...
+    "exportedFiles": {
+      "description": "List of files that should be exported",
+      "type": "array",
+      "uniqueItems": true,
+      "items": {
+        "type": "object",
+        "$ref": "exportedFile.json"
+      }
+    },
+    "jobProfile": {
```
VictorSoroka1: Can we name it the same way it is named in ui-data-import? [jobProfileInfo](https://github.com/folio-org/data-import-raml-storage/blob/master/schemas/mod-source-record-manager/jobExecution.json#L27), so UI should be able to be reused both for data-export and data-import.

IlliaDaliek: Yes, sure

VictorSoroka1: Thanks. By the way, will the `name` field be added to the `jobProfileInfo`? I see for data-import there is a such and it is used on the UI. It would be great for our case to have the same.

```
jobProfileInfo: {
  id: '32fafcc3-f582-493d-88b0-3c538480cd83',
  name: 'Standard BIB profile',
},
```

IlliaDaliek: Okay, will add it now

ramls/schemas/jobExecution.json
Igor-Gorchakov: The information about user's first name and last name can not be gathered from metadata on UI, let's have a separate field for this

PR #21, opened 2020-03-06
Mdexp 58 - bugfix

mrdpopov:
Purpose
   * At this moment ExportManager and InputDataManager transfer data with each other directly, using created instances. We should enable data-transfer based on an event bus.
    https://vertx.io/docs/vertx-service-proxy/java/

Approach
  * Update InitApi: initialize(Vertx.createProxy) and use generated proxies instead of component instances


Igor-Gorchakov:
Thanks for creating the PR, we need this to allow Managers tp to communicate to e
each other on an event bus

KVupp:
Thanks for the context @Igor-Gorchakov . Please add it to the description of the PR so it is clear to everyone

PR #26, opened 2020-03-11
MDEXP-67

Igor-Gorchakov:
## Purpose
The marc records are temporarily stored in local storage. Once the retrieval from SRS is completed, the created file must be exported to S3. Create export method for performing the operation

## Approach
Once the last chunk of UUIDs are retrieved and written to the file, the file must be exported to S3


src/main/java/org/folio/service/export/storage/AWSStorageServiceImpl.java
```diff
@@ -0,0 +1,63 @@
...
+    String bucketName = getProperty("bucket.name");
+    if (StringUtils.isEmpty(bucketName)) {
+      throw new IllegalStateException("S3 bucket name is not defined. Please set the bucket.name system property");
+    } else {
+      TransferManager transferManager = amazonFactory.getTransferManager();
+      try {
+        LOGGER.info("Uploading generated binary file {} to bucket {}", fileDefinition, bucketName);
+        MultipleFileUpload multipleFileUpload = transferManager.uploadDirectory(
+          bucketName,
+          PARENT_FOLDER,
```
Igor-Gorchakov: ALL: in this place we just save the file into a predefined folder. This should be reworked later to follow _/tenantId/jobId/fileId/fileName.marc_

src/main/java/org/folio/service/export/LocalFileSystemExportService.java
```diff
@@ -1,7 +1,14 @@
 package org.folio.service.export;
 
+import java.io.ByteArrayInputStream;
```
KVupp: Let's use a common formatter going forward, so that the PRs are cleaner and import order need not change

src/main/java/org/folio/service/export/storage/AWSStorageServiceImpl.java
KVupp: we will need it for the release, let's please change to use this format

KVupp: You can just construct the key name from those values

src/main/java/org/folio/service/export/LocalFileSystemExportService.java
mrdpopov: This is a common import formating, which was used to format all the files. What do you think should be changed?

Igor-Gorchakov: This is the default IDE setting for imports, we should configure IDEs to use the same order of imports. Let's do this next week, no time to waste

KVupp: yea, that's the reason I mentioned going forward. We can leave it as-is for now

src/main/java/org/folio/service/export/storage/AWSStorageServiceImpl.java
```diff
@@ -0,0 +1,64 @@
...
+   * @return A link using which the file can be downloaded
+   */
+  @Override
+  public String getFileDownloadLink(String jobId, String exportFileId, String tenantId) {
+    return StringUtils.EMPTY;
+  }
+
+  @Override
+  public void storeFile(FileDefinition fileDefinition, String tenantId) {
+    String parentFolder = String.format("%s/%s/%s.%s", tenantId, fileDefinition.getJobExecutionId(), fileDefinition.getId(), MARC_FILE_EXTENSION);
```
KVupp: shouldn't the extension be at the end of the file name? no at the parent folder level

src/main/java/org/folio/service/manager/exportmanager/ExportManagerImpl.java
```diff
@@ -78,15 +78,14 @@ protected void exportBlocking(ExportPayload exportPayload) {
...
-
     SrsLoadResult srsLoadResult = loadSrsMarcRecordsInPartitions(identifiers, params);
     exportService.export(srsLoadResult.getUnderlyingMarcRecords(), fileExportDefinition);
     List<JsonObject> instances = loadInventoryInstancesInPartitions(srsLoadResult.getInstanceIdsWithoutSrs(), params);
     List<String> mappedMarcRecords = mappingService.map(instances);
     exportService.export(mappedMarcRecords, fileExportDefinition);
 
     if (exportPayload.isLast()) {
-      exportService.postExport(fileExportDefinition);
+      exportService.postExport(fileExportDefinition, params.getTenantId());
```
KVupp: what if the postExport fails? There should be some exception handling to mark it as error?

src/test/resources/mockData/generatedBinaryFile.txt
```diff
@@ -0,0 +1 @@
+EMPTY FILE THAT IS GENERATED WHILE EXPORT FOR THE TESTING NEEDS
```
KVupp: I know this is just a mock file, but please use allowable file extensions and formats.

src/main/java/org/folio/service/export/storage/AWSStorageServiceImpl.java
KVupp: Also, don't have the fileDefinition Id in the path, while fetching the file , we'll just have access to jobExecution details

KVupp: so it has to be just tenantId/jobExecutionID/fileName.mrc

mrdpopov: The intention was to just copy all the files previously generated. There is a jira to address the missing extension. https://issues.folio.org/browse/MDEXP-72

src/main/java/org/folio/service/manager/exportmanager/ExportManagerImpl.java
mrdpopov: It is there already at a higher level. See ExportManagerImpl:expotData

src/test/resources/mockData/generatedBinaryFile.txt
mrdpopov: Sure.

src/main/java/org/folio/service/export/storage/AmazonFactory.java
```diff
@@ -0,0 +1,22 @@
...
+
+import com.amazonaws.regions.Regions;
+import com.amazonaws.services.s3.AmazonS3;
+import com.amazonaws.services.s3.AmazonS3ClientBuilder;
+import com.amazonaws.services.s3.transfer.TransferManager;
+import com.amazonaws.services.s3.transfer.TransferManagerBuilder;
+
+@Component
+public class AmazonFactory {
+  private Regions region = Regions.DEFAULT_REGION;
```
KVupp: this shouldn't be default region, please remove it. The region will be pulled from the environment as well.

src/main/java/org/folio/service/export/storage/AWSStorageServiceImpl.java
KVupp: but here it's the parent folder string. If I'm not wrong the above STring will be "tenatid/jobExeId/fileDefId.mrc" which is incorrect isn;t it?

src/main/java/org/folio/service/manager/exportmanager/ExportManagerImpl.java
Igor-Gorchakov: If _postExport_  throws an exception, then it triggers WorkerExecutor to call _ExportManagerImpl.handleExportResult_ with failed asyncResult

src/main/java/org/folio/service/export/storage/AWSStorageServiceImpl.java
Igor-Gorchakov: fixed

src/main/java/org/folio/service/export/storage/AmazonFactory.java
Igor-Gorchakov: @KVupp, I agree. Could you please fix it in your branch after merging this PR

Igor-Gorchakov:
I've fixed all the critical places/comments. I think it is ready to merge now

src/main/java/org/folio/service/export/storage/AmazonFactory.java
KVupp: Will do

Igor-Gorchakov:
I pushed the last commit and going to merge it

pom.xml
```diff
@@ -3,7 +3,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>org.folio</groupId>
   <artifactId>mod-data-export</artifactId>
-  <version>1.0.0-SNAPSHOT</version>
+  <version>1.0.0-SNAPSHOT-1</version>
```
KVupp: does it need a suffix?

mrdpopov: Fixed already, sorry for a typo

src/main/java/org/folio/service/export/storage/AWSStorageServiceImpl.java
```diff
@@ -74,13 +83,32 @@ private Date getExpiration() {
...
 
+
   @Override
-  public void storeFile() {
-    // Will be implemented as part of MDEXP-67
+  public void storeFile(FileDefinition fileDefinition, String tenantId) {
+    String folderInS3 = tenantId + "/" + fileDefinition.getJobExecutionId();
+    String bucketName = getProperty(BUCKET_PROP_KEY);
+    if (StringUtils.isNullOrEmpty(bucketName)) {
+      throw new IllegalStateException("S3 bucket name is not defined. Please set the bucket.name system property");
```
KVupp: I think it's better to be consistent in throwing this error, change it here or in the above method

mrdpopov: Sure.

PR #27, opened 2020-03-11
MDEXP-69: Update a jobExecution entry in DB

IlliaDaliek:
## Purpose
Update status of job execution in database in pair with file definition export process.

### DIFF
--- a/src/main/java/org/folio/dao/impl/JobExecutionDaoImpl.java
+++ b/src/main/java/org/folio/dao/impl/JobExecutionDaoImpl.java
@@ -25,7 +25,7 @@ import org.folio.rest.persist.interfaces.Results;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Repository;
 
-import static org.folio.dao.impl.FileDefinitionDaoImpl.constructCriteria;
+import static org.folio.util.HelperUtils.constructCriteria;
 
 @Repository
 public class JobExecutionDaoImpl implements JobExecutionDao {
--- a/src/main/java/org/folio/dao/impl/JobExecutionDaoImpl.java
+++ b/src/main/java/org/folio/dao/impl/JobExecutionDaoImpl.java
@@ -97,7 +97,7 @@ public class JobExecutionDaoImpl implements JobExecutionDao {
     }
     return promise.future()
       .map(Results::getResults)
-      .map(fileDefinitions -> fileDefinitions.isEmpty() ? Optional.empty() : Optional.of(fileDefinitions.get(0)));
+      .map(jobExecutions -> jobExecutions.isEmpty() ? Optional.empty() : Optional.of(jobExecutions.get(0)));
   }
 
   /**
--- a/src/main/java/org/folio/dao/impl/JobExecutionDaoImpl.java
+++ b/src/main/java/org/folio/dao/impl/JobExecutionDaoImpl.java
@@ -1,11 +1,16 @@
 package org.folio.dao.impl;
 
-import static org.folio.util.HelperUtils.constructCriteria;
-
 import io.vertx.core.Future;
 import io.vertx.core.Promise;
 import io.vertx.core.logging.Logger;
 import io.vertx.core.logging.LoggerFactory;
+
+import javax.ws.rs.NotFoundException;
+
+import java.lang.invoke.MethodHandles;
+import java.util.Optional;
+import java.util.UUID;
+
 import org.folio.cql2pgjson.CQL2PgJSON;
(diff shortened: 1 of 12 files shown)


src/main/java/org/folio/service/manager/inputdatamanager/InputDataManagerImpl.java
```diff
@@ -82,24 +88,63 @@ protected void initBlocking(JsonObject request, JsonObject params) {
...
+  /**
+   * Update jobExecution status with IN-PROGRESS && update exported files && update completed date
+   *
+   * @param jobExecution         job execution
+   * @param fileExportDefinition definition of the file to export
+   * @param tenantId             tenant id
+   */
+  private void updateJobExecutionStatusAndExportedFiles(JobExecution jobExecution, FileDefinition fileExportDefinition, String tenantId) {
+    ExportedFile exportedFile = new ExportedFile()
+      .withFileId(fileExportDefinition.getId())
```
KVupp: we shouldn't be re-using the UUIDs, let;s have a unique Id for fileId

src/main/java/org/folio/dao/impl/JobExecutionDaoImpl.java
```diff
@@ -49,13 +59,52 @@
...
+  public Future<JobExecution> update(JobExecution jobExecution, String tenantId) {
+    Promise<JobExecution> promise = Promise.promise();
+    try {
+      Criteria idCrit = constructCriteria(ID_FIELD, jobExecution.getId());
+      pgClientFactory.getInstance(tenantId).update(TABLE, jobExecution, new Criterion(idCrit), true, updateResult -> {
+        if (updateResult.failed()) {
+          LOGGER.error("Could not update jobExecution with id {}", jobExecution.getId(), updateResult.cause());
+          promise.fail(updateResult.cause());
+        } else if (updateResult.result().getUpdated() != 1) {
+          String errorMessage = String.format("JobExecution with id '%s' was not found", jobExecution.getId());
```
KVupp: Please change to use appropriate error message

src/main/java/org/folio/service/manager/inputdatamanager/InputDataManagerImpl.java
```diff
@@ -82,24 +87,69 @@ protected void initBlocking(JsonObject request, JsonObject params) {
...
+    jobExecutionService.getById(id, tenantId).compose(optionalJobExecution -> {
+      optionalJobExecution.map(jobExecution -> {
+        ExportedFile exportedFile = new ExportedFile()
+          .withFileId(UUID.randomUUID().toString())
+          .withFileName(fileExportDefinition.getFileName());
+        Set<ExportedFile> exportedFiles = jobExecution.getExportedFiles();
+        exportedFiles.add(exportedFile);
+        jobExecution.setExportedFiles(exportedFiles);
+        jobExecution.setStatus(JobExecution.Status.IN_PROGRESS);
+        jobExecution.setCompletedDate(new Date());
```
KVupp: why are we setting the completed date, when the job is still in progress?

src/main/java/org/folio/service/manager/inputdatamanager/InputDataManagerImpl.java
IlliaDaliek: Thank you for note, i will fix it

src/main/java/org/folio/service/manager/inputdatamanager/InputDataManagerImpl.java
IlliaDaliek: You are right, It is unnecessary here, will remove it

src/main/java/org/folio/dao/impl/JobExecutionDaoImpl.java
```diff
@@ -58,27 +62,51 @@
...
+      Criteria idCrit = constructCriteria(ID_FIELD, jobExecutionId);
       pgClientFactory.getInstance(tenantId).get(TABLE, JobExecution.class, new Criterion(idCrit), false, promise);
     } catch (Exception e) {
       LOGGER.error(e);
       promise.fail(e);
     }
     return promise.future()
       .map(Results::getResults)
-      .map(jobExecutions -> jobExecutions.isEmpty() ? Optional.empty() : Optional.of(jobExecutions.get(0)));
+      .map(fileDefinitions -> fileDefinitions.isEmpty() ? Optional.empty() : Optional.of(fileDefinitions.get(0)));
```
KVupp: isn't this job executions?

IlliaDaliek: I've already corrected that, thank you

PR #30, opened 2020-03-12
hotfixes

mrdpopov:


KVupp:
this was one of the other observations too, /export returning a success even if FileDefinition is invalid. Thanks for creating this, but it would help if you added some context in the PR description too

PR #33, opened 2020-03-17
MDEXP-76 Fix NPE if none of the records are present in SRS

andrii-novytskyi:
## Purpose
https://issues.folio.org/browse/MDEXP-76

## Approach
Add validation to avoid NPE when the uploaded input file has instance IDs , for which none of them have SRS records.
Covered with unit tests.


src/main/java/org/folio/service/export/LocalFileSystemExportService.java
```diff
@@ -55,6 +55,13 @@ public void export(List<String> jsonRecords, FileDefinition fileDefinition) {
 
   @Override
   public void postExport(FileDefinition fileDefinition, String tenantId) {
-    exportStorageService.storeFile(fileDefinition, tenantId);
+    if (isValidFileDefinition(fileDefinition)) {
+      exportStorageService.storeFile(fileDefinition, tenantId);
+    }
+  }
+
```
Igor-Gorchakov: Add javadoc

src/test/java/org/folio/service/export/ExportServiceUnitTest.java
```diff
@@ -55,10 +55,21 @@ public void export_shouldPassExportFor_NULL_Records() {
...
     // then
     Mockito.verify(exportStorageService, Mockito.times(1)).storeFile(any(FileDefinition.class), anyString());
   }
+
+  public void postExport_shouldNotStoreFileFor_Null_SourcePath() {
+    // given
+    FileDefinition fileDefinition = new FileDefinition()
+      .withSourcePath(null);
+    // when
+    exportService.postExport(fileDefinition, "tenant");
```
Igor-Gorchakov: move "tenant" into constant field

src/test/java/org/folio/service/export/ExportServiceUnitTest.java
```diff
@@ -55,10 +55,21 @@ public void export_shouldPassExportFor_NULL_Records() {
...
+    // then
+    Mockito.verify(exportStorageService, Mockito.times(0)).storeFile(any(FileDefinition.class), anyString());
```
Igor-Gorchakov: use Mockito.never() instead of times(0)

src/main/java/org/folio/service/export/LocalFileSystemExportService.java
andrii-novytskyi: The javadoc is present in the method contract, should I add it to the implementation as well?

src/test/java/org/folio/service/export/ExportServiceUnitTest.java
```diff
@@ -55,10 +57,30 @@ public void export_shouldPassExportFor_NULL_Records() {
...
-    FileDefinition fileDefinition = new FileDefinition();
+    FileDefinition fileDefinition = new FileDefinition()
+      .withSourcePath("files/mockData/generatedBinaryFile.mrc");
     // when
-    exportService.postExport(fileDefinition, "tenant");
+    exportService.postExport(fileDefinition, TENANT);
     // then
     Mockito.verify(exportStorageService, Mockito.times(1)).storeFile(any(FileDefinition.class), anyString());
   }
+
```
Igor-Gorchakov: @Test

src/test/java/org/folio/service/export/ExportServiceUnitTest.java
```diff
@@ -55,10 +57,30 @@ public void export_shouldPassExportFor_NULL_Records() {
...
+  public void postExport_shouldNotStoreFileFor_Null_FileDefinition() {
+    // given
+    FileDefinition fileDefinition = null;
+    // when
+    exportService.postExport(fileDefinition, TENANT);
+    // then
+    Mockito.verify(exportStorageService, Mockito.never()).storeFile(any(FileDefinition.class), anyString());
+  }
+
```
Igor-Gorchakov: @test

Igor-Gorchakov:
Yes, the job should be definitely failed in this case. 
We can just create a class for ExportException with a cause (enum), similar to HttpException.
The cause will be NO_RECORDS_FOUND, BUCKET_IS_NOT_SPECIFIED, CAN_NOT_CREATE_BUCKET and so on. Once the exception is thrown, then InputDataManager handles the exception- updates the job status depending on the cause.

KVupp:
> We can just create a class for ExportException with a 'cause' field(enum).
I already created HttpException class for the same reason, so that we can reuse it to throw our module specific exceptions. we can just add another ENUM in the error codes class.

Igor-Gorchakov:
> > We can just create a class for ExportException with a 'cause' field(enum).
> 
> I already created HttpException class for the same reason, so that we can reuse it to throw our module specific exceptions. we can just add another ENUM in the error codes class.

Okay, then let's rename HttpException to _ServiceException_ or _ExportException_ to make it more generic. The case when there are no records to export is not 'Http' case

src/main/java/org/folio/service/manager/export/ExportResult.java
```diff
@@ -1,5 +1,86 @@
...
-  COMPLETED, IN_PROGRESS, ERROR;
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import io.vertx.codegen.annotations.DataObject;
+import io.vertx.core.json.JsonObject;
+import org.folio.util.ErrorCode;
+
+@DataObject(generateConverter = true)
+@JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
+public class ExportResult {
```
Igor-Gorchakov: There is a discussion held about how to pass an error code from ExportManager to InputDataManager in order to let the InputDataManager update job with error cause.
_@DataObject_ annotation is used to make ExportResult be serializable on an event bus.

Igor-Gorchakov:
@KVupp  waiting for your review

src/main/java/org/folio/rest/exceptions/ServiceException.java
```diff
@@ -0,0 +1,48 @@
...
+    this.message = StringUtils.EMPTY;
+  }
+
+  public int getCode() {
+    return status.toInt();
+  }
+
+  public Error getError() {
+    return StringUtils.isEmpty(getMessage())
+      ? new Error().withCode(errorCode.getCode()).withMessage(errorCode.getDescription())
```
KVupp: If you assign the message directly above in line 26. This additional empty check is not needed

src/main/java/org/folio/service/manager/export/ExportResult.java
```diff
@@ -1,5 +1,86 @@
...
+
+  public ExportResult(JsonObject jsonObject) {
+    ExportResultConverter.fromJson(jsonObject, this);
+  }
+
+  public static ExportResult completed() {
+    return COMPLETED;
+  }
+
+  public static ExportResult inProgress() {
```
KVupp: Can this class be made enum, and leave the methods as is? Not sure how the dataobject works, but just checking

src/main/java/org/folio/rest/exceptions/ServiceException.java
andrii-novytskyi: This is a check which constructor was used to create a ServiceException, to return a valid value from getError() method

src/main/java/org/folio/service/manager/export/ExportResult.java
andrii-novytskyi: I am not sure, but i think this is more informative solution, than simple enum.
As it is impossible in vert.x to make this class serializable on an event bus (it is possible for enum or json format, for example), so this annotation helps us to introduce ExportResult object in json format, and then remap it again to the object.

Igor-Gorchakov: This class can not be an enum, because enum can not hold the state. Enum fields are final static, but we need to hold the state of error, to pass error code back to the InputDataManager

PR #34, opened 2020-03-18
MDEXP-71 Tests covering end-end flow

illia-borysenko:
## Overview

The stories for Export Manager https://issues.folio.org/browse/MDEXP-58 and File upload is complete, it is better to have tests to cover the end-end flow .
Now InputDataManager & ExportManager are integrated into the master branch.
We should cover the communication between InputDataManager & ExportManager with a unit test.

## Testing details

- cover with test cases changing the number of UUIDs in CSV file
- assert that the marc binary file is generated and saved locally
- assert name of the generated marc file
- cover with negative cases: empty CSV file, CSV file containing UUIDs in the wrong format

## Acceptance Criteria

- Unit tests covering both positive and negative scenarios for end-end flow are written


src/test/java/org/folio/rest/impl/ExportManagerTest.java
```diff
@@ -5,108 +5,367 @@
...
-import static org.hamcrest.Matchers.is;
-import static org.hamcrest.Matchers.notNullValue;
+import static org.folio.rest.jaxrs.model.JobExecution.Status.SUCCESS;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.doNothing;
+import static org.mockito.Mockito.when;
 
 @RunWith(VertxUnitRunner.class)
 public class ExportManagerTest extends RestVerticleTestBase {
```
Igor-Gorchakov: Let's keep the ExportManagerTest untouched, better create a new class for end-to-end test and write tests there

src/test/java/org/folio/rest/impl/EndToEndTest.java
```diff
@@ -0,0 +1,353 @@
...
+    //when
+    ExportRequest exportRequest = getExportRequest(uploadedFileDefinition);
+    RestAssured.given()
+      .spec(jsonRequestSpecification)
+      .body(JsonObject.mapFrom(exportRequest).encode())
+      .when()
+      .post(EXPORT_URL);
+
+    // then
+    vertx.setTimer(1000L, handler -> {
```
KVupp: I don't think using a timer is a good idea. There is a chance that the tests can fail in other environments

illia-borysenko: I don't know other options, without timer test will fail definitely, the same approach is used for mod-inventory. It's possible to use Awaitlity library, but it can take some time to investigate how to use it in our cases

illia-borysenko: Maybe at least we can set delay longer

PR #39, opened 2020-03-20
MDEXP-74 Add file extenstion valiation to /data-export/fileDefinitions

illia-borysenko:
## Description
Currently the validation happens on the UI but the validation should occur on the backend as well in order to prevent any malicious files being uploaded.

## Acceptance criteria

- validation only for /data-export/fileDefinitions
- Support .csv extension
- API tests are added

### DIFF
--- a/src/main/java/org/folio/rest/impl/DataExportImplFileDefinitionImpl.java
+++ b/src/main/java/org/folio/rest/impl/DataExportImplFileDefinitionImpl.java
@@ -113,7 +113,7 @@ public class DataExportImplFileDefinitionImpl implements DataExportFileDefinitio
 
   private Future<Void> validateFileNameExtension(String fileName) {
     if (!FilenameUtils.isExtension(fileName.toLowerCase(), CSV_FORMAT_EXTENSION)) {
-      throw new ServiceException(HttpStatus.HTTP_UNPROCESSABLE_ENTITY, "File name extension does not corresponds csv format");
+      throw new ServiceException(HttpStatus.HTTP_UNPROCESSABLE_ENTITY, ErrorCode.INVALID_UPLOADED_FILE_EXTENSION);
     }
     return succeededFuture();
   }
--- a/src/main/java/org/folio/rest/impl/DataExportImplFileDefinitionImpl.java
+++ b/src/main/java/org/folio/rest/impl/DataExportImplFileDefinitionImpl.java
@@ -1,19 +1,11 @@
 package org.folio.rest.impl;
 
-import static io.vertx.core.Future.succeededFuture;
-import static org.folio.rest.RestVerticle.STREAM_ABORT;
-import static org.folio.util.ExceptionToResponseMapper.map;
-import static org.folio.rest.jaxrs.model.FileDefinition.Status;
-
-
 import io.vertx.core.AsyncResult;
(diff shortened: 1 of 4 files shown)


src/main/java/org/folio/rest/impl/DataExportImplFileDefinitionImpl.java
```diff
@@ -107,4 +110,11 @@ public void getDataExportFileDefinitionsByFileDefinitionId(String fileDefinition
       .setHandler(asyncResultHandler);
 
   }
+
+  private Future<Void> validateFileNameExtension(String fileName) {
+    if (!FilenameUtils.isExtension(fileName.toLowerCase(), CSV_FORMAT_EXTENSION)) {
+      throw new ServiceException(HttpStatus.HTTP_UNPROCESSABLE_ENTITY, "File name extension does not corresponds csv format");
```
KVupp: Can you add this to the ErrorCode enum class, so that we have all error messages in a single place

src/test/java/org/folio/rest/impl/FileUploadServiceTest.java
```diff
@@ -62,6 +62,24 @@ public void postFileDefinition_return200Status(TestContext context) {
...
+      .withId(UUID.randomUUID().toString())
+      .withFileName("InventoryUUIDs.txt");
+    // when created a new entity
+    Response response = RestAssured.given()
+      .spec(jsonRequestSpecification)
+      .body(JsonObject.mapFrom(givenFileDefinition).encode())
+      .when()
+      .post(FILE_DEFINITION_SERVICE_URL);
+    // then retrieve it and verify
+    context.assertEquals(HttpStatus.SC_UNPROCESSABLE_ENTITY, response.getStatusCode());
```
KVupp: Also verify the message returned is the same one that is being thrown, so we know it doesn't fail with a different error

src/main/java/org/folio/rest/impl/DataExportImplFileDefinitionImpl.java
```diff
@@ -22,15 +14,25 @@
...
 import org.folio.rest.tools.utils.TenantTool;
-import org.folio.service.file.upload.FileUploadService;
 import org.folio.service.file.definition.FileDefinitionService;
+import org.folio.service.file.upload.FileUploadService;
 import org.folio.spring.SpringContextUtil;
 import org.folio.util.ErrorCode;
 import org.folio.util.ExceptionToResponseMapper;
 import org.springframework.beans.factory.annotation.Autowired;
 
+import javax.ws.rs.core.Response;
```
KVupp: Let's discuss about the common code formatter next time we meet, so that we don't have unnecessary changes in PR. 
No changes necessary now

src/test/java/org/folio/rest/impl/FileUploadServiceTest.java
illia-borysenko: done

src/main/java/org/folio/rest/impl/DataExportImplFileDefinitionImpl.java
illia-borysenko: done

PR #40, opened 2020-03-21
MDEXP-76 Fix empty error message

andrii-novytskyi:
## Purpose
Fix empty error message


andrii-novytskyi:
Logs with the correct error message.
picturemessage 2ii

Igor-Gorchakov:
AFAIK there are other places in the project where getCause() is logged. Please, fix those places as well

PR #41, opened 2020-03-26
MDEXP-80 Provide Number of Records

illia-borysenko:
## Description
Once the export jobs complete, the number of records should be made available to UI so that the landing page - log of completed jobs can be populated correctly. The number of records should be a number of records that were successfully exported.


src/main/java/org/folio/service/job/JobExecutionServiceImpl.java
```diff
@@ -38,4 +42,23 @@
...
+  public Future<JobExecution> incrementCurrentProgress(final String jobExecutionId, final int delta, final String tenantId ) {
+        return jobExecutionDao.getById(jobExecutionId, tenantId)
+          .compose(jobExecutionOptional ->{
+            if(jobExecutionOptional.isPresent()) {
+              JobExecution jobExecution = jobExecutionOptional.get();
+              Progress progress = jobExecution.getProgress();
+              if(nonNull(progress)) {
+                int current = nonNull(progress.getCurrent()) ? progress.getCurrent() : 0;
+                int incrementedCurrent = current + delta;
+                progress.setCurrent(incrementedCurrent);
```
KVupp: It's good that we are updating the progress. But where do we plan to update the total records(progress.total)?
That is where the UI will pickup the number right?

illia-borysenko: I think one of the options is to set total records in InputDataManager, if we're able to get all numbers of UUIDs at once then we can do it during init, otherwise will collect and increase the total number each time when reading next chunk of UUIDs

KVupp: On the last export chunk, we can just copy in progress to total records right? That will be straight forward I believe

Igor-Gorchakov: @KVupp   in my vision the _total_ identifies a number of all the UUIDs are in the CSV file. We can put it before the export started, just in the _init_ method of the InputDataManager

Igor-Gorchakov:
@illia-borysenko  please remove the HttpServerTestBase, that is duplicated (org.folio.rest.impl)

PR #52, opened 2020-04-09
MDEXP-92 Change naming of the file generated by export

andrii-novytskyi:
## Purpose
https://issues.folio.org/browse/MDEXP-92

## Approach
- In the naming of the file generated by export, HRID is used instead of timestamp.

### DIFF
--- a/src/main/java/org/folio/rest/impl/DataExportHelper.java
+++ b/src/main/java/org/folio/rest/impl/DataExportHelper.java
@@ -1,9 +1,9 @@
 package org.folio.rest.impl;
 
 import io.vertx.core.Future;
+import org.apache.commons.lang3.StringUtils;
 import org.folio.HttpStatus;
 import org.folio.rest.exceptions.ServiceException;
-import org.folio.rest.jaxrs.model.ExportedFile;
 import org.folio.rest.jaxrs.model.FileDownload;
 import org.folio.service.export.storage.ExportStorageService;
 import org.folio.service.job.JobExecutionService;
@@ -28,15 +28,19 @@ public class DataExportHelper {
 
   private Future<String> getDownloadFileName(String jobExecutionId, String exportFileId, String tenantId) {
     return jobExecutionService.getById(jobExecutionId, tenantId)
-      .map(job -> job.map(jb -> jb.getExportedFiles()
+      .map(job -> job.getExportedFiles()
         .stream()
         .filter(expFile -> expFile.getFileId()
           .equals(exportFileId))
-        .findFirst())
+        .findFirst()
         .orElseThrow(() -> new ServiceException(HttpStatus.HTTP_NOT_FOUND, String.format("Job with id: %s not found", jobExecutionId))))
-      .map(exportedFile -> exportedFile.map(ExportedFile::getFileName)
-          .orElseThrow(() -> new ServiceException(HttpStatus.HTTP_NOT_FOUND, String.format("Export File with id: %s not found: ", exportFileId))));
-
+      .compose(exportedFile -> {
+        if (StringUtils.isEmpty(exportedFile.getFileName())) {
+          throw new ServiceException(HttpStatus.HTTP_NOT_FOUND, String.format("Export File with id: %s not found: ", exportFileId));
+        } else {
+          return Future.succeededFuture(exportedFile.getFileName());
+        }
+      });
   }
 
 }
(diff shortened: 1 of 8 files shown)


andrii-novytskyi:
image

Verified on local env that file contains HRID.

src/main/java/org/folio/rest/impl/DataExportHelper.java
```diff
@@ -28,15 +28,19 @@
...
         .stream()
         .filter(expFile -> expFile.getFileId()
           .equals(exportFileId))
-        .findFirst())
+        .findFirst()
         .orElseThrow(() -> new ServiceException(HttpStatus.HTTP_NOT_FOUND, String.format("Job with id: %s not found", jobExecutionId))))
-      .map(exportedFile -> exportedFile.map(ExportedFile::getFileName)
-          .orElseThrow(() -> new ServiceException(HttpStatus.HTTP_NOT_FOUND, String.format("Export File with id: %s not found: ", exportFileId))));
-
+      .compose(exportedFile -> {
```
KVupp, line 37: I know the above line was not introduced in this PR, but I think if the job is not found then it will be a failed future thrown by getById method, it doesn't come to line 36. So , the error message should belong to file  not found

src/main/java/org/folio/rest/impl/DataExportHelper.java
```diff
@@ -28,15 +28,19 @@
...
+        if (StringUtils.isEmpty(exportedFile.getFileName())) {
```
KVupp, line 38: fileName is a mandatory field in schema, this check is not required

src/main/java/org/folio/service/job/JobExecutionServiceImpl.java
```diff
@@ -49,25 +53,31 @@
...
+          return failedFuture(new NotFoundException(errorMessage));
+        }
+      });
   }
 
   @Override
   public void updateJobStatusById(String id, JobExecution.Status status, String tenantId) {
-    getById(id, tenantId).compose(optionalJobExecution -> {
-      optionalJobExecution.ifPresent(jobExecution -> {
+    getById(id, tenantId).onSuccess(jobExecution -> {
```
KVupp, line 71: what happens if the update fails?

KVupp:
Please make sure code coverage is above 80%

src/main/java/org/folio/service/job/JobExecutionServiceImpl.java
andrii-novytskyi: @KVupp we just logging the error on the DAO level

src/main/java/org/folio/rest/impl/DataExportHelper.java
andrii-novytskyi: removed this check, thank you

src/main/java/org/folio/service/manager/input/InputDataManagerImpl.java
```diff
@@ -92,34 +89,35 @@ protected void initBlocking(JsonObject request, Map<String, String> params) {
...
+              }
+            });
+          } else {
+            fileDefinitionService.save(fileExportDefinition.withStatus(FileDefinition.Status.ERROR), tenantId);
+            jobExecutionService.updateJobStatusById(jobExecutionId, JobExecution.Status.FAIL, tenantId);
+            sourceReader.close();
+          }
         } else {
-          fileDefinitionService.save(fileExportDefinition.withStatus(FileDefinition.Status.ERROR), tenantId);
+          LOGGER.error(String.format("Failed to start export process, file definition status with id %s is not COMPLETED", exportRequest.getFileDefinitionId()));
```
KVupp, line 117: If the export is triggered before the file upload is completed, then the job execution need not fail. WE just can throw wrror on the export API and allow them to retry. 
With the current logic, we'll have to upload the entire file again to continue

KVupp: Sorry to add some comments which were not introduced in this PR, but We shouldn't have missed them in earlier stories.
And we need to correct them somewhere. If you think this is more work than expected, please create a bug and link it to this PR. And then we can close this PR

andrii-novytskyi: @KVupp according to the endpoint contract, we always return 204 status code, so I don't think we are able to do it. However, we are getting job execution before checking the file definition status. It seems that checking file definition status not needed here because we will fail if we will not found job execution, what do you think? Should we remove this check?

Igor-Gorchakov: It's not possible to start export before file has been uploaded, because the job gets created on upload completion. No need to check file definition status, it's enough to check if the job exists

PR #53, opened 2020-04-13
MDEXP-116 Move PoC to module sources

Igor-Gorchakov:
https://issues.folio.org/browse/MDEXP-116

## Purpose
Move sources from personal repo to Folio repo

## Approach
- Copied sources
- Added mapping for 245 '0' '0' index title to let Magda to verify it
- Renamed _rules.json_  to _rulesDeprecated.json_,  created _rulesDefault.json_ to keep relevant version there
- Added unit tests to cover translation function
- Verified by MarcValidator
- Tested locally, it works :)

## Pre-Merge Checklist


src/main/java/org/folio/service/mapping/processor/RuleProcessor.java
```diff
@@ -0,0 +1,105 @@
...
+ * @see EntityReader
+ * @see TranslationFunction
+ * @see RecordWriter
+ */
+public final class RuleProcessor {
+  private List<Rule> rules = new ArrayList<>();
+
+  public RuleProcessor(JsonArray rules) {
+    Iterator ruleIterator = rules.iterator();
+    while (ruleIterator.hasNext()) {
```
andrii-novytskyi, line 42: Do we really need iterator and while cycle here? I guess we can use for each

src/main/java/org/folio/service/mapping/writer/impl/AbstractRecordWriter.java
```diff
@@ -0,0 +1,103 @@
...
+  protected abstract void writeDataField(RecordDataField recordDataField);
+
+  private RecordDataField buildDataFieldForListOfStrings(ListValue listValue) {
+    DataSource dataSource = listValue.getDataSource();
+    String tag = listValue.getDataSource().getTag();
+    RecordDataField field = new RecordDataField(tag);
+    for (String stringValue : listValue.getValue()) {
+      if (listValue.getDataSource().isSubFieldSource()) {
+        char subFieldCode = dataSource.getSubField().charAt(0);
+        String subFieldData = stringValue;
```
andrii-novytskyi, line 67: looks like "subFieldData" is redundant here

src/main/java/org/folio/service/mapping/processor/rule/Rule.java
```diff
@@ -0,0 +1,26 @@
...
+import io.vertx.core.json.JsonObject;
+
+import java.util.ArrayList;
+import java.util.List;
+
+
+public class Rule {
+  private List<DataSource> dataSources = new ArrayList<>();
+
+  public Rule(JsonObject rule) {
```
Igor-Gorchakov, line 13: @KVupp    this converting logic from JSON object to POJO  is customized. There is a need to pass tag to datasource, I don't know how to convert json object in other was

src/main/java/org/folio/service/mapping/processor/RuleProcessor.java
```diff
@@ -0,0 +1,103 @@
...
+      Rule rule = new Rule(JsonObject.mapFrom(ruleIterator.next()));
+      this.rules.add(rule);
+    }
+  }
+
+  public String process(EntityReader reader, RecordWriter writer, Settings settings) {
+    this.rules.forEach(rule -> {
+      RuleValue ruleValue = reader.read(rule);
+      switch (ruleValue.getType()) {
```
illia-borysenko, line 51: The idea for refactoring - implement design pattern Command instead of using a switch operator. For example, you can create two classes SimpleRuleProcessor and CompositeRuleProcessor and move there translate(...) methods.  Static final Map can be used o initialize and use such methods

src/main/java/org/folio/service/mapping/writer/impl/AbstractRecordWriter.java
```diff
@@ -0,0 +1,103 @@
...
+    String tag = listValue.getDataSource().getTag();
+    RecordDataField field = new RecordDataField(tag);
+    for (String stringValue : listValue.getValue()) {
+      if (listValue.getDataSource().isSubFieldSource()) {
+        char subFieldCode = dataSource.getSubfield().charAt(0);
+        String subFieldData = stringValue;
+        field.addSubField(subFieldCode, subFieldData);
+      } else if (dataSource.isIndicatorSource()) {
+        char indicator = stringValue.charAt(0);
+        if ("1".equals(dataSource.getIndicator())) {
```
illia-borysenko, line 71: (minor) Let's extract indicator values to constants

src/main/java/org/folio/service/mapping/processor/RuleProcessor.java
Igor-Gorchakov: Good idea but let's keep it simple at this moment and leave as it is

src/test/resources/mapping/given_inventory_instance.json
```diff
@@ -0,0 +1,3 @@
+{
+  "title": "The Journal of ecclesiastical history"
```
Igor-Gorchakov, line 2: Once we have more mapping rules, we will expand this instance with more new fields, that are affected by mapping

src/test/resources/processor/given_entity.json
```diff
@@ -0,0 +1,31 @@
+{
```
KVupp, line 1: instead of an arbitrary json, it would be good to use the real instance

src/main/java/org/folio/service/mapping/writer/impl/AbstractRecordWriter.java
```diff
@@ -0,0 +1,105 @@
...
+          writeControlField(recordControlField);
+        }
+      }
+    }
+  }
+
+  @Override
+  public void write(CompositeValue compositeValue) {
+    String tag = compositeValue.getValue().get(0).get(0).getDataSource().getTag();
+    for (List<StringValue> entry : compositeValue.getValue()) {
```
KVupp, line 52: I think we should call the  write(SimpleValue simpleValue)  here for each value, because each type can then have string or array types isn't it?

src/main/java/org/folio/service/mapping/writer/impl/JsonRecordWriter.java
```diff
@@ -0,0 +1,18 @@
+package org.folio.service.mapping.writer.impl;
+
+import org.marc4j.MarcJsonWriter;
+import org.marc4j.MarcWriter;
+
+import java.io.ByteArrayOutputStream;
+import java.io.OutputStream;
+
+public class JsonRecordWriter extends MarcRecordWriter {
```
KVupp, line 9: The hierarcy seems off here, why does a JsonREcordWriter extend a MarcREcordWriter?

src/main/java/org/folio/service/mapping/writer/impl/AbstractRecordWriter.java
```diff
@@ -0,0 +1,105 @@
...
+import org.folio.service.mapping.writer.fields.RecordControlField;
+import org.folio.service.mapping.writer.fields.RecordDataField;
+
+import java.util.Collections;
+import java.util.List;
+
+import static org.folio.service.mapping.reader.values.SimpleValue.SubType.LIST_OF_STRING;
+import static org.folio.service.mapping.reader.values.SimpleValue.SubType.STRING;
+
+public abstract class AbstractRecordWriter implements RecordWriter {
```
KVupp, line 18: I don't think we need both abstract class and interface, we can create default methods in the interface and move the method implementations to RecordWriter, so that all the other types can implement just one Interface

src/main/java/org/folio/service/mapping/processor/rule/Rule.java
KVupp: I added a branch to see how it can be made a little simpler: https://github.com/folio-org/mod-data-export/pull/55

src/main/java/org/folio/service/mapping/processor/rule/DataSource.java
```diff
@@ -0,0 +1,49 @@
+package org.folio.service.mapping.processor.rule;
+
+import io.vertx.core.json.JsonObject;
+
+public class DataSource {
```
KVupp, line 5: It would be good if all these classes are moved to a json , and let the POJOs be generated

src/test/resources/processor/given_entity.json
Igor-Gorchakov: This entity is just an arbitrary JSON to show various kinds of rules for the RuleProcessorTest :)   The real instance is in the _given_inventory_instance.json_ file, see comment above :)

src/main/java/org/folio/service/mapping/writer/impl/JsonRecordWriter.java
Igor-Gorchakov: The hierarchy is right is this case, because both XmlRecordWriter and JsonRecordWriter extend marc4j mechanisms to build record :)

src/main/java/org/folio/service/mapping/writer/impl/AbstractRecordWriter.java
Igor-Gorchakov: The purpose of an interface is just to provide a contract to a caller, no need to overload interface with extra logic :)

src/main/java/org/folio/service/mapping/processor/rule/DataSource.java
Igor-Gorchakov: The benefit of using schemas for POJO generating is just to provide support for marshalling/unmarshalling from POJO to request/response. We don't need such a benefit here. In addition, all this staff about records generating will be moved to a shared library without RMB-related things.   So, I don't think it is a good idea to use POJO generating here :)

src/main/java/org/folio/service/mapping/writer/impl/AbstractRecordWriter.java
Igor-Gorchakov: Yes, we can, but lets this logic separated. Some of these places may be changed after rework in future

src/main/java/org/folio/service/mapping/processor/rule/DataSource.java
Igor-Gorchakov: I saw the PR in a separated branch, looks good

PR #57, opened 2020-04-17
MDEXP-93 Generate MARC bib record containing 001, 245 and 999ff fields

illia-borysenko:
## Purpose
When the user provides Instance UUID and there is no underlying SRS record we will create most basic MARC bib record that contains only three fields 001, 245 $a and 999 ff $i in accordance with the recommended mapping:

- instance.json field hrid is mapped to 001
- instance.json field title is mapped to 245 00 $a
- instance.json filed id is mapped to 999 ff $i

Since the records will be generated in parallel to retrieving records from SRS the formatting should not require any addition data manipulation by the consumer of the data and is formatted consistently with parsedRecord.json.

## Acceptance criteria
1. The generated record is short-lived and generated each time it is requested
2. The record is saved in .mrc file and contains 001, 245$a 999ff$i fields are present and populated with the correct data
3. Unit tests are written


src/test/resources/mapping/expected_marc_record.mrc
```diff
@@ -1 +1 @@
-00080nam a2200037 a 450024500420000000aThe Journal of ecclesiastical history
\ No newline at end of file
+00147nam a2200061 a 4500001000200000245004200002999004100044100aThe Journal of ecclesiastical historyffic8b50e3f-0446-429c-960e-03774b88223f
```
KVupp, line 1: Looks like this is an invalid marc record
Screen Shot 2020-04-21 at 9 32 26 AM

illia-borysenko: I will check, thanks

KVupp: looks good now

KVupp:
Looks like we pointed the story a lot higher :)

PR #60, opened 2020-04-27
MDEXP-102 Generate MARC bib record - 655 $a/natureOfContentId

Igor-Gorchakov:
## Purpose
MDEXP-102 Generate MARC bib record - NatureOfContentTerm, replacing keys with values
## Approach

- Applied caching for Inventory Setting by the Job id. Caffeine cache is the best lightweight among other opensource implementations. It provides both blocking and async implementations with a list of additional configurations. Using expiration strategy for selfcleaning - 60 seconds after last access by the given jobId.
https://github.com/ben-manes/caffeine
https://www.baeldung.com/java-caching-caffeine

- Updated rulesDefault.json for 655 field $a subfield, ind1 is blank, ind2 equals 4

- Updated and factored unit tests for HTTP clients

image

## Pre-Merge Checklist


Igor-Gorchakov:
> I did not go through the entire code but from the description and "MappingSettingsProvider " class looks like the cache is being stored by job id, without taking tenantId into consideration.
> In a multi tenant environment this will not work because multiple tenants will have same job id and the settings will be different for each one of them

Agree, job id is tenant-specific itself, because the database schema belongs for some tenant

src/main/java/org/folio/service/mapping/settings/MappingSettingsProvider.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.List;
+
+/**
+ * The component is responsible to provide settings and parameters needed for mapping.
+ * Caches settings for the given job id.
+ * If requested settings are not found by the given job id then performs loading.
+ */
+@Component
+public class MappingSettingsProvider {
+  private static final int CACHE_EXPIRATION_AFTER_ACCESS_SECONDS = 60;
```
KVupp, line 19: Currently, for a medium sized load the job is running for 2-3 hours. Given that I think we can just do cache eviction when the job has ended right? because if it is  a minute for medium sized jobs we are still fetching all the entries per read

Igor-Gorchakov: The entries do not get expired every 60 seconds, because no one chunk of uuids does not take more than 60 seconds of processing.
It is about to expire entries after the last access, that will be 2-3 hours + 60 seconds in this case

src/main/java/org/folio/clients/ClientUtil.java
```diff
@@ -1,23 +1,78 @@
...
 import static org.folio.rest.RestVerticle.OKAPI_HEADER_TENANT;
 import static org.folio.rest.RestVerticle.OKAPI_HEADER_TOKEN;
 
 public final class ClientUtil {
+  private static final Logger LOGGER = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
   private ClientUtil() {
   }
 
+  public static Optional<JsonObject> getByIds(List<String> ids, OkapiConnectionParams params, String pattern, String queryPattern) {
```
andrii-novytskyi, line 37: To make it more readable, maybe we should change the name of "pattern" variable  to uriPattern or something like that

src/main/java/org/folio/clients/ClientUtil.java
```diff
@@ -1,23 +1,78 @@
...
+
   public static void setCommonHeaders(HttpRequestBase requestBase, OkapiConnectionParams params) {
     requestBase.setHeader(OKAPI_HEADER_TOKEN, params.getToken());
     requestBase.setHeader(OKAPI_HEADER_TENANT, params.getTenantId());
     requestBase.setHeader(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_JSON);
     requestBase.setHeader((HttpHeaders.ACCEPT), MediaType.APPLICATION_JSON);
   }
+
+  @NotNull
+  private static URI prepareFullUri(List<String> ids, OkapiConnectionParams params, String recordPattern, String queryPattern) {
```
andrii-novytskyi, line 57: The same for "recordPattern"

src/main/java/org/folio/clients/InventoryClient.java
```diff
@@ -0,0 +1,62 @@
...
+
+  private static final int SETTING_LIMIT = 100;
+  private static final String NATURE_OF_CONTENT_TERMS_URL = "%s/nature-of-content-terms?limit=" + SETTING_LIMIT;
+
+  public Optional<JsonObject> getInstancesByIds(List<String> ids, OkapiConnectionParams params, int partitionSize) {
+    return ClientUtil.getByIds(ids, params, GET_INSTANCES_URL + LIMIT_PATTERN + partitionSize, QUERY_PATTERN_INVENTORY);
+  }
+
+  public List<JsonObject> getNatureOfContentTerms(OkapiConnectionParams params) {
+    return getSettingsByUrl(NATURE_OF_CONTENT_TERMS_URL, params, "natureOfContentTerms");
```
andrii-novytskyi, line 39: I think "natureOfContentTerms" could be static final variable, but it is not a problem if you will leave it as it is now

src/main/java/org/folio/service/mapping/processor/translations/Settings.java
```diff
@@ -1,4 +1,18 @@
 package org.folio.service.mapping.processor.translations;
```
andrii-novytskyi, line 1: It is better to move this class to the settings folder

src/main/java/org/folio/service/mapping/processor/translations/TranslationsHolder.java
```diff
@@ -1,16 +1,43 @@
...
+    public String apply(String id, JsonObject parameters, Settings settings) {
+      Optional<JsonObject> optionalEntry = settings.getNatureOfContentTerms().stream()
+        .filter(entry -> id.equals(entry.getString(ID)))
+        .findFirst();
+      if (optionalEntry.isPresent()) {
+        JsonObject entry = optionalEntry.get();
+        return entry.getString("name");
+      } else {
+        LOGGER.info("Nature of content term is not found by the given id: {}", id);
+        return null;
```
illia-borysenko, line 30: Maybe let's use an empty string value instead of null?

src/main/java/org/folio/clients/ClientUtil.java
```diff
@@ -1,23 +1,78 @@
...
+    }
+  }
+
+  private static JsonObject getResponseEntity(CloseableHttpResponse response) {
+    HttpEntity entity = response.getEntity();
+    if (response.getStatusLine().getStatusCode() == HttpStatus.SC_OK && entity != null) {
+      try {
+        return new JsonObject(EntityUtils.toString(entity));
+      } catch (IOException e) {
+        LOGGER.error("Exception while requesting instances", e);
```
KVupp, line 73: Please add proper log message, it's not always fetching instances right?

src/main/java/org/folio/clients/InventoryClient.java
```diff
@@ -0,0 +1,63 @@
...
+import static java.lang.String.format;
+
+@Component
+public class InventoryClient {
+  private static final Logger LOGGER = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+  private static final String GET_INSTANCES_URL = "%s/instance-storage/instances?query=(%s)";
+  private static final String QUERY_PATTERN_INVENTORY = "id==%s";
+  private static final String LIMIT_PATTERN = "&limit=";
+
+  private static final int SETTING_LIMIT = 100;
```
KVupp, line 31: the UI uses 200 to fetch the records to show on UI, let's use the same for consistency

src/main/java/org/folio/service/mapping/processor/translations/TranslationsHolder.java
```diff
@@ -1,16 +1,44 @@
...
+    @Override
+    public String apply(String id, JsonObject parameters, Settings settings) {
+      Optional<JsonObject> optionalEntry = settings.getNatureOfContentTerms().stream()
+        .filter(entry -> id.equals(entry.getString(ID)))
+        .findFirst();
+      if (optionalEntry.isPresent()) {
+        JsonObject entry = optionalEntry.get();
+        return entry.getString("name");
+      } else {
+        LOGGER.info("Nature of content term is not found by the given id: {}", id);
```
KVupp, line 30: should be logged as an error

src/main/java/org/folio/service/mapping/settings/MappingSettingsProvider.java
KVupp: I got that point that it expires after 60 seconds of the last access, it doesn't work if the UUIDs are intertwined , say 100 entries of records in SRS followed by 100 entries to generate records on the fly. 
Anyway not a huge deal , but let's try to keep an eye on  cache misses if required in the future.

src/main/java/org/folio/service/mapping/processor/translations/TranslationsHolder.java
```diff
@@ -1,16 +1,44 @@
...
   SET_VALUE() {
     @Override
     public String apply(String value, JsonObject parameters, Settings settings) {
       return parameters.getString("value");
     }
+  },
+  SET_NATURE_OF_CONTENT_TERM() {
+    @Override
+    public String apply(String id, JsonObject parameters, Settings settings) {
+      Optional<JsonObject> optionalEntry = settings.getNatureOfContentTerms().stream()
```
KVupp, line 23: Probably makes sense to use a Map to store the settings with Id as the key , instead of a list, instead of traversing it each time, map gives you a constant time operation
https://github.com/folio-org/mod-data-export/pull/60/files#diff-f16bbac4236a4ff04032f19d7b0776fcR43

src/test/java/org/folio/service/mapping/MappingServiceUnitTest.java
```diff
@@ -15,21 +22,41 @@
...
 public class MappingServiceUnitTest {
 
+  @Mock
+  private MappingSettingsProvider mappingSettingsProvider;
+  private String jobExecutionId = "67429e0e-601a-423b-9a29-dec4a30c8534";
+  private OkapiConnectionParams params = new OkapiConnectionParams();
+
+  @Before
+  public void mockSettings() {
+    Settings settings = new Settings();
```
KVupp, line 32: Why can't you just use the existing mock file. This is redundant isn't it, and any change needs to be made it multiple places.

src/test/java/org/folio/service/mapping/TranslationFunctionUnitTest.java
```diff
@@ -1,27 +1,63 @@
...
     Assert.assertEquals("new field value", result);
   }
+
+  @Test
+  public void SetNatureOfContentTerm_shouldReturnTermName() {
+    // given
+    TranslationFunction translationFunction = TranslationsHolder.lookup("set_nature_of_content_term");
+    String value = "96879b60-098b-453b-bf9a-c47866f1ab2a";
+    Settings settings = new Settings();
+    JsonObject natureOfContentTerm_audioBook = new JsonObject()
```
KVupp, line 36: same here, just reuse the existing mock file

src/main/resources/rules/rulesDefault.json
```diff
@@ -63,5 +63,27 @@
         "from": "$.hrid"
       }
     ]
+  },
+  {
+    "tag": "655",
```
KVupp, line 68: can we keep the tags sorted please, it'll be easier to track which one's are supported just by looking at the  file

src/test/java/org/folio/service/mapping/MappingServiceUnitTest.java
```diff
@@ -15,21 +22,41 @@
...
+    JsonObject natureOfContentTerm_audioBook = new JsonObject()
+      .put("id", "96879b60-098b-453b-bf9a-c47866f1ab2a")
+      .put("name", "audiobook")
+      .put("source", "folio");
+    JsonObject natureOfContentTerm_autobiography = new JsonObject()
+      .put("id", "04a6a8d2-f902-4774-b15f-d8bd885dc804")
+      .put("name", "autobiography")
+      .put("source", "folio");
+    settings.addNatureOfContentTerms(Arrays.asList(natureOfContentTerm_audioBook, natureOfContentTerm_autobiography));
+    Mockito.when(mappingSettingsProvider.getSettings(jobExecutionId, params)).thenReturn(settings);
```
KVupp, line 42: My only concern with this that there will not be an end-end test if we are mocking this part. Can we just setup vertx mocks, instead of mocking the call. I think it is already done above in testing the client class, would be better to just use them here

descriptors/ModuleDescriptor-template.json
```diff
@@ -31,7 +31,8 @@
           "modulePermissions": [
             "users.item.get",
             "source-storage.records.get",
-            "inventory-storage.instances.collection.get"
+            "inventory-storage.instances.collection.get",
+            "inventory-storage.nature-of-content-terms.collection.get"
```
KVupp, line 35: need to add interface "nature-of-content-terms" 1.0 in the requires field

src/test/java/org/folio/service/mapping/MappingServiceUnitTest.java
Igor-Gorchakov: This is just a unit test, it is not related to end-to-end testing

Igor-Gorchakov: The idea of unit test is to cover only one piece of code - MappingService, no need to cover remote calls. We have tests for http clients to cover remote calls with vertx mocks

Igor-Gorchakov: Also, I plan to create a user story to rewrite EndToEnd test to use Vertx mocks

KVupp: Exactly, that's my point, we should not be adding separate stories to cover tests. It is a good idea to cover all the tests here

> The idea of unit test is to cover only one piece of code - 
I know what a unit test is Igor :) . My point is that we should be writing end-end-tests here, it's not much of an effort because we already have the code for it in the class that I mentioned

PR #63, opened 2020-05-03
MDEXP-149 - Instances with quotes not being processed

KVupp:
https://issues.folio.org/browse/MDEXP-149

## Purpose
It was observed that Instances with underlying SRS record were also being generated with instances on the fly. Upon investigation it was found that if the UUIDs were enclosed in "", they were not being removed from the Hashset thereby considering that they were already fetched from SRS.

## Approach
- Remove quotes when reading from file
- Refactoring all the APIs to a single class for easy retrieval
 
Additional
- Added Integration test cases, where Mockserver is setup that makes calls to External APIs to fetch json records
- Observed few issues along the way in unit tests, capturing them in TODO

#### TODOS and Open Questions
- [ ] Observed that the EndToEnd tests use csv files with "," That format of csv file is not being exported correctly
- [ ] Nature of Content Terms is being called even if there are no records to be generated on the fly
- [x] Currently the EndToEnd test mocks all Clients and uses Spring context to load those beans , but that is preventing from running Integration tests, where they should not be mocked. (Temporarily commented EndToEndTests but need to fix this issue)
- [ ] Any failure in EndToEnd tests is not failing the test run, but instead gets timeout with exception 
```
org.mockito.exceptions.base.MockitoException: 
No argument value was captured!
You might have forgotten to use argument.capture() in verify()...
...or you used capture() in stubbing but stubbed method was not called.
Be aware that it is recommended to use capture() only with verify()

Examples of correct argument capturing:
    ArgumentCaptor<Person> argument = ArgumentCaptor.forClass(Person.class);
    verify(mock).doSomething(argument.capture());
    assertEquals("John", argument.getValue().getName());
```

## Pre-Merge Checklist


src/test/java/org/folio/rest/impl/EndToEndTest.java
```diff
@@ -62,318 +59,297 @@
 
 @RunWith(VertxUnitRunner.class)
 public class EndToEndTest extends RestVerticleTestBase {
```
illia-borysenko, line 61: Let's just ignore test instead of commenting as it was discussed

KVupp:
There are a few other TODOs listed  in this PR. I will capture them in a story in the backlog

PR #65, opened 2020-05-04
MDEXP-144 POC: Handle mapping transformations in Mapping profile

illia-borysenko:
## Overview
When user specifies mapping transformations while creating mapping profile, they need to be considered while exporting records:
Note: For Q2, only holdings and items transformations will be present
The scope of this POC is to cover the below scenarios:
1) If holdings and item transformations are given in mapping profile, how will the "from" field be stored for these values
2) How will the list of values be displayed on the UI
3) How will these transformation be combined with default mapping rules for Instances
4) How can sorting of the fields be handled
5) How can these transformations be used to modify MARC records fetched from SRS
6) Can this approach be reused for Instance transformations later? or a new approach is needed?


src/main/java/org/folio/service/mapping/profiles/MappingProfileService.java
```diff
@@ -0,0 +1,97 @@
...
+        List<MappingProfileTransformation> settingsFieldNames = buildTransformationsBySettings(recordType, okapiConnectionParams, transformationConfig);
+        fieldNames.addAll(settingsFieldNames);
+      } else {
+        fieldNames.add(buildSimpleTransformation(recordType, transformationConfig));
+      }
+    }
+    return fieldNames;
+  }
+
+  private List<MappingProfileTransformation> buildTransformationsBySettings(RecordType recordType, OkapiConnectionParams okapiConnectionParams, TransformationConfig transformationConfig) {
```
KVupp, line 72: I confirmed with Magda today, URL Relationship is not needed, so we can skip all the implementation  for Q2. WE can keep this PR though for further reference. We can just simply return the fixed values

src/main/java/org/folio/service/mapping/profiles/TransformationConfig.java
```diff
@@ -0,0 +1,71 @@
...
+import org.folio.service.mapping.profiles.loader.ElectronicAccessRelationshipsLoader;
+import org.folio.service.mapping.profiles.loader.SettingsLoader;
+
+public enum TransformationConfig {
+
+  PERMANENT_LOCATION_ID("permanentLocationId",
+                        "$.{recordType}[*].permanentLocationId",
+                        "Permanent location"),
+  MATERIAL_TYPE_ID("materialTypeId",
+                   "$.{recordType}[*].materialTypeId",
```
KVupp, line 12: Material type is just a string field , not an array. Need not change it here , but just FYI for implementation

illia-borysenko: Array sign is set only for recordType but not for materialTypeId, for item it will look $.item[*].materialTypeId

KVupp:
The implementation stories corresponding to this PR have been completed

PR #71, opened 2020-05-13
MDEXP-132 Get the total number of records to be exported

illia-borysenko:
## Description
In order to display the progress bar for the export jobs in progress on the data export landing page we need a total count of the records requested for the export (should be equal to the number of UUIDs in the input file.

## Acceptance criteria

The total number of the requested records is returned
The number is correct
Unit tests are written
API tests are written


ramls/schemas/jobExecution.json
```diff
@@ -44,7 +44,7 @@
         },
         "total": {
           "description": "Total number of records",
-          "type": "integer"
+          "type": "string"
```
Igor-Gorchakov, line 47: update samples

PR #75, opened 2020-05-18
MDEXP-131 Create default mapping profile

andrii-novytskyi:
## Purpose

https://issues.folio.org/browse/MDEXP-131

## Approach
Created SQL script with default mapping profile entity, that runs during the module deployment.

## Pre-Merge Checklist


src/main/java/org/folio/rest/impl/ModTenantAPI.java
```diff
@@ -52,11 +66,41 @@ private void executeStorageCleanUpJob(Vertx vertx, OkapiConnectionParams params)
...
+
+      String tenantId = TenantTool.calculateTenantId(headers.get(X_OKAPI_TENANT));
+      String moduleName = PostgresClient.getModuleName();
+
+      sqlScript = sqlScript.replace(TENANT_PLACEHOLDER, tenantId).replace(MODULE_PLACEHOLDER, moduleName);
+
+      Promise<List<String>> promise = Promise.promise();
+      PostgresClient.getInstance(context.owner()).runSQLFile(sqlScript, false, promise);
+
+      LOGGER.info("Module is being deployed in test mode, test data will be initialized. Check the server log for details.");
```
KVupp, line 98: What is test mode?

andrii-novytskyi: Updated the message

src/main/java/org/folio/rest/impl/ModTenantAPI.java
```diff
@@ -52,11 +66,41 @@ private void executeStorageCleanUpJob(Vertx vertx, OkapiConnectionParams params)
...
+    try {
+      InputStream inputStream = this.getClass().getClassLoader().getResourceAsStream(script);
+
+      if (inputStream == null) {
+        LOGGER.info("Default data was not initialized: no resources found: {}", script);
+        return Future.succeededFuture();
+      }
+
+      String sqlScript = IOUtils.toString(inputStream, StandardCharsets.UTF_8.name());
+      if (StringUtils.isBlank(sqlScript)) {
```
KVupp, line 86: Is there a reason for using sql script and not TenantLoading in RMB? 
That is much cleaner and easier to use than script, that way data is in json format

andrii-novytskyi: Changed to use TenantLoading

src/main/java/org/folio/rest/impl/ModTenantAPI.java
```diff
@@ -17,10 +19,13 @@
 import java.util.Map;
 
 public class ModTenantAPI extends TenantAPI {
+  private static final Logger LOGGER = LoggerFactory.getLogger(ModTenantAPI.class);
 
   private static final long DELAY_TIME_BETWEEN_CLEANUP_VALUE_MILLIS = 3600_000;
-
-  private static final Logger logger = LoggerFactory.getLogger(ModTenantAPI.class);
+  private static final String LOAD_SAMPLE_PARAMETER = "loadSample";
```
KVupp, line 25: As this is default data, let's call it that, and not sample

src/main/resources/data/mappingProfiles/defaultMappingProfiles.json
```diff
@@ -0,0 +1,20 @@
...
+  ],
+  "outputFormat": "MARC",
+  "userInfo": {
+    "firstName": "DIKU",
+    "lastName": "ADMINISTRATOR",
+    "userName": "diku_admin"
+  },
+  "metadata": {
+    "createdDate": "2020-05-15T09:39:13.659+0000",
+    "createdByUserId": "08d3ea27-e4ff-4a26-83e1-d5f86f16495c",
```
KVupp, line 16: Let's not add the userId, it will not be available in all the environments

src/main/resources/data/mappingProfiles/defaultMappingProfiles.json
```diff
@@ -0,0 +1,20 @@
+{
+  "id": "25d81cbe-9686-11ea-bb37-0242ac130002",
+  "name": "Default instance mapping profile",
+  "description": "Default mapping profile for the inventory instance record",
+  "recordTypes": [
+    "INSTANCE"
+  ],
+  "outputFormat": "MARC",
+  "userInfo": {
+    "firstName": "DIKU",
```
KVupp, line 10: Same here, let's call this System process or something, there won't be a diku_admin in all environments

src/main/resources/data/mappingProfiles/defaultMappingProfiles.json
andrii-novytskyi: @KVupp looks like there is a validation in RMB, and an empty string is not pass through the regex

andrii-novytskyi: i will remove createdByUserId and updatedByUserId fields from metadata then

src/main/java/org/folio/rest/impl/ModTenantAPI.java
andrii-novytskyi: sure, i will rename it

src/main/java/org/folio/rest/impl/ModTenantAPI.java
```diff
@@ -31,13 +36,25 @@ public ModTenantAPI() { //NOSONAR
...
   @Override
-  public void postTenant(TenantAttributes entity, Map<String, String> headers, Handler<AsyncResult<Response>> handlers, Context context)  {
+  public void postTenant(TenantAttributes entity, Map<String, String> headers, Handler<AsyncResult<Response>> handlers, Context context) {
     super.postTenant(entity, headers, asyncResult -> {
       if (asyncResult.failed()) {
         handlers.handle(asyncResult);
       } else {
         initStorageCleanupService(headers, context);
-        handlers.handle(asyncResult);
+        TenantLoading tenantLoading = new TenantLoading();
```
illia-borysenko, line 45: Move all new code to a separate private method

src/test/java/org/folio/rest/impl/EntitiesCrudTest.java
```diff
@@ -38,11 +39,19 @@
   @EnumSource(TestEntities.class)
   void testVerifyCollection(TestEntities testEntity) throws MalformedURLException {
     logger.info(String.format("--- mod-data-export %s test: Verifying database's initial state ... ", testEntity.name()));
-    getRequest(testEntity.getEndpoint()).then()
-      .log()
-      .all()
-      .statusCode(200)
-      .body("totalRecords", equalTo(0));
+    if (testEntity.getClazz().getName().equals(MappingProfile.class.getName())) {
```
KVupp, line 42: Though this is not a good way to do it, I think this will change once we add default job profile too, so we can ignore for this PR

andrii-novytskyi: I planned to move it back in PR about job profile, but ok, i will ignore them

PR #83, opened 2020-05-27
MDEXP-162 change trasnfromations.enabled to boolean type

illia-borysenko:
## Purpose
Change type of a field transformations.enabled to boolean

## Pre-Merge Checklist

### DIFF
--- a/ramls/schemas/profiles/transformations.json
+++ b/ramls/schemas/profiles/transformations.json
@@ -15,7 +15,7 @@
     },
     "enabled": {
       "description": "Indicates whether field should be mapped",
-      "type": "string"
+      "type": "boolean"
     },
     "transformation": {
       "description": "Mapping expression",
--- a/ramls/schemas/profiles/transformations.json
+++ b/ramls/schemas/profiles/transformations.json
@@ -15,7 +15,8 @@
     },
     "enabled": {
       "description": "Indicates whether field should be mapped",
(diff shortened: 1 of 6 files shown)


ramls/schemas/profiles/transformations.json
```diff
@@ -15,7 +15,7 @@
     },
     "enabled": {
       "description": "Indicates whether field should be mapped",
-      "type": "string"
+      "type": "boolean"
```
KVupp, line 18: I think also adding default false would make sense

src/test/resources/profiles/mappingProfile.json
```diff
@@ -11,7 +11,7 @@
       "fieldId": "electronicAccess.linkText",
       "path": "Json path for electronicAccess[*].linkText",
```
KVupp, line 12: Is there a reason for using the text in path field? I remember it was not supported in sample file because of raml issues. But we should probably be using proper format in tests

ramls/schemas/profiles/transformations.json
illia-borysenko: Thanks, it's good point, I will add default false

src/test/resources/profiles/mappingProfile.json
illia-borysenko: Updated

PR #84, opened 2020-05-28
MDEXP-162 Mapping profile - transformations - process holdings and items simple fields

illia-borysenko:
## Description
As a librarian exporting inventory data I would like to have records contain the electronic access data so that items information is appended to inventory instance records when exporting in the MARC format

## Scenario 1:
Given: any field associated with the item electronic access has associated transformation
When: the inventory instance record has an inventory item record with populated electronic access data
Then: the data that is stored in the item record is used to populate fields as specified in the transformation and the MARC fields are appended to MARC bib record

## Scenario 2:
Given: any field associated with the item electronic access has associated transformation
When: the inventory instance record has an inventory item record without electronic access data
Then: the mapping is skipped and no fields are appended to the MARC bib record

Note: User story covers transformation of the fields:
Holdings - Call number - prefix
Holdings - Call number
Holdings - Call number - suffix
Holdings - Electronic access - Link text
Holdings - Electronic access - URI
Items - Effective call number
Items - Electronic access - Link text
Items - Electronic access - URI

## Additional Acceptance criteria:
Unit tests are written

## Pre-Merge Checklist


src/main/java/org/folio/service/mapping/processor/RuleFactory.java
```diff
@@ -1,56 +1,119 @@
...
+    List<DataSource> dataSources = new ArrayList<>();
+    DataSource fromDataSource = new DataSource();
+    fromDataSource.setFrom(mappingTransformation.getPath());
+    dataSources.add(fromDataSource);
+    String transformation = mappingTransformation.getTransformation();
+    Pattern pattern = Pattern.compile(SUBFIELD_REGEX);
+    Matcher matcher = pattern.matcher(transformation);
+    if (matcher.find()) {
+      fromDataSource.setSubfield(matcher.group());
+      dataSources.add(buildEmptyIndicatorDataSource(INDICATOR_1));
```
KVupp, line 100: Did we get a conformation that the indicators will to be always blank? What if the user specifies an indicator like 1 or 2

illia-borysenko: Updated

src/main/java/org/folio/service/mapping/processor/RuleFactory.java
```diff
@@ -1,56 +1,120 @@
...
+  }
+
+  private List<DataSource> buildDataSources(Transformations mappingTransformation) {
+    List<DataSource> dataSources = new ArrayList<>();
+    DataSource fromDataSource = new DataSource();
+    fromDataSource.setFrom(mappingTransformation.getPath());
+    dataSources.add(fromDataSource);
+    String transformation = mappingTransformation.getTransformation();
+    Pattern pattern = Pattern.compile(SUBFIELD_REGEX);
+    Matcher matcher = pattern.matcher(transformation);
```
KVupp, line 96: Isn't the last character the subfield, why can't you just do a substring there as well? Why do you need a pattern matcher

src/main/java/org/folio/service/manager/export/ExportManagerImpl.java
```diff
@@ -111,22 +108,24 @@ protected void exportBlocking(ExportPayload exportPayload) {
...
       instanceWithHoldingsAndItems.put("instance", instance);
-      List<JsonObject> holdings = recordLoaderService.getHoldingsForInstance(instance.getString("id"), params);
-      instanceWithHoldingsAndItems.put("holdings", new JsonArray(holdings));
-      List<String> holdingIds = holdings.stream()
-        .map(jo -> jo.getString("id"))
-        .collect(Collectors.toList());
-      List<JsonObject> items = recordLoaderService.getAllItemsForHolding(holdingIds, params);
-      instanceWithHoldingsAndItems.put("items", new JsonArray(items));
-
+      if(recordTypes.contains(RecordType.HOLDINGS) || recordTypes.contains(RecordType.ITEM)) {
```
KVupp, line 116: I think it would be good to also check if the transformations are present here. So that these are no called unecessarily. (currently there is nothing preveting us from seleting record types and not adding transformations in the UI)

src/test/java/org/folio/service/mapping/MappingServiceUnitTest.java
```diff
@@ -91,5 +96,45 @@ void shouldMapInstance_to_marcRecord() {
...
+    // then
+    Assert.assertEquals(1, actualMarcRecords.size());
+    String actualMarcRecord = actualMarcRecords.get(0);
+    String expectedMarcRecord = readFileContentFromResources("mapping/expected_marc_record_with_holdings_and_items.mrc");
+    Assert.assertEquals(expectedMarcRecord, actualMarcRecord);
+  }
+
+  private List<Transformations> createHoldingsAndItemSimpleFieldTransformations() {
+    List<Transformations> transformations = new ArrayList<>();
+    transformations.add(createTransformations("callNumber", "$.holdings[*].callNumber", "900  $a", HOLDINGS));
```
KVupp, line 119: Can you please add atleast one scenario of different indicators instead of just blanks, so that everything is tested

src/main/java/org/folio/service/mapping/processor/RuleFactory.java
illia-borysenko: I think in future we will support multiple subfields in one transformation value, therefore it's better to parse it by regex than by position. At the same time inc case if a user adds a lot of spaces it will not fail

src/main/java/org/folio/service/manager/export/ExportManagerImpl.java
illia-borysenko: Ok, good point, I will add this verification

src/test/java/org/folio/service/mapping/MappingServiceUnitTest.java
illia-borysenko: Agree, will add one more test case with indicators

src/test/java/org/folio/service/mapping/processor/RuleFactoryTest.java
```diff
@@ -0,0 +1,238 @@
...
+  private static final String TRANSFORMATION_FIELD_VALUE_WITH_SUBFIELD = "002 $a";
+  private static final String TRANSFORMATIONS_PATH_2 = "transformationsPath2";
+  private static final String TRANSFORMATION_FIELD_VALUE_2 = "003";
+  private static final String SUBFIELD_A = "a";
+  private static final String FIRST_INDICATOR = "1";
+  private static final String SET_VALUE_FUNCTION = "set_value";
+  private static final String VALUE_PARAMETER = "value";
+  private static final String SECOND_INDICATOR = "2";
+
+  @InjectMocks
```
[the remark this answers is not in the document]
illia-borysenko: Yeah, you are right, thanks

PR #94, opened 2020-06-05
MDEXP-189 Fix NPE for 008 field

andrii-novytskyi:
## Purpose
https://issues.folio.org/browse/MDEXP-189

## Approach
Added checks for fields that are filled in the field '008'.

## Pre-Merge Checklist

### DIFF
--- a/src/main/java/org/folio/service/mapping/processor/translations/TranslationsHolder.java
+++ b/src/main/java/org/folio/service/mapping/processor/translations/TranslationsHolder.java
@@ -128,9 +128,9 @@ public enum TranslationsHolder implements TranslationFunction {
   SET_FIXED_LENGTH_DATA_ELEMENTS() {
     private transient DateTimeFormatter originCreatedDateFormatter = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSSZ");
     private transient DateTimeFormatter targetCreatedDateFormatter = DateTimeFormatter.ofPattern("yyMMdd");
-    private String datesOfPublicationPattern = "datesOfPublication";
-    private String languagesPattern = "languages";
-    private String fieldPattern = "%s|%s%s||||||||       |||||%s||";
+    private static final String DATES_OF_PUBLICATION = "datesOfPublication";
+    private static final String LANGUAGES = "languages";
+    private static final String FIELD_PATTERN = "%s|%s%s||||||||       |||||%s||";
 
     @Override
     public String apply(String originCreatedDate, int currentIndex, Translation translation, ReferenceData referenceData, Metadata metadata) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/folio/service/mapping/processor/translations/TranslationsHolder.java
```diff
@@ -126,16 +128,29 @@ public String apply(String updatedDate, int currentIndex, Translation translatio
   SET_FIXED_LENGTH_DATA_ELEMENTS() {
     private transient DateTimeFormatter originCreatedDateFormatter = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSSZ");
     private transient DateTimeFormatter targetCreatedDateFormatter = DateTimeFormatter.ofPattern("yyMMdd");
+    private String datesOfPublicationPattern = "datesOfPublication";
```
KVupp, line 131: these are constants, please use the proper naming, with capitals and final for more readability

andrii-novytskyi: sure, I will push changes now

src/main/java/org/folio/service/mapping/processor/translations/TranslationsHolder.java
```diff
@@ -126,16 +128,29 @@ public String apply(String updatedDate, int currentIndex, Translation translatio
...
+      } else {
+        createdDateParam = targetCreatedDateFormatter.format(ZonedDateTime.now());
+      }
 
       String publicationDate0Param = "||||";
       String publicationDate1Param = "||||";
-      if (metadata != null && metadata.getData().containsKey("datesOfPublication")) {
-        List<String> publicationDates = (List<String>) metadata.getData().get("datesOfPublication").getData();
+      if (metadata != null && metadata.getData().containsKey(DATES_OF_PUBLICATION)
+        && metadata.getData().get(DATES_OF_PUBLICATION) != null) {
```
KVupp, line 152: I think something is incorrect here:
The publication field looks like this:
```
"publication" : [ {
      "publisher" : "American Bar Association",
      "place" : "Chicago, Ill.",
      "dateOfPublication" : null,
      "role" : null
    } ],
```

look at "dateOfPublication". There is no "date**s**OfPublication". 
Also, it returns an array, you should be checking if each entry is null or empty

KVupp: to be more clear, what if the metadata contains:   metadata.addData("datesOfPublication",  [null, 2016]);. Is this case handled?

andrii-novytskyi: agree, not handled., will add more checks

PR #97, opened 2020-06-12
MDEXP-178 - Release 2.0.0

KVupp:
https://issues.folio.org/browse/MDEXP-178
## Purpose
 Q2 2020 release 

## Approach
- change the versions of ramls with breaking changes in schema(item-storage)
- bump up dependant interface versions in module descriptor

#### TODOS and Open Questions
- [x] should it be 1.2.0 or 2.0.0 given the breaking changes in schema


ramls/dataExport-FileDefinition.raml
```diff
@@ -1,7 +1,7 @@
 #%RAML 1.0
 
 title: Data export
-version: v1.0
+version: v2.0
```
andrii-novytskyi, line 4: Don`t we need to do up version in profiles.raml?

PR #102, opened 2020-06-18
MDEXP-194 Provide information from jobProfile to jobExecution

andrii-novytskyi:
## Purpose
https://issues.folio.org/browse/MDEXP-194

## Approach
1.During the export process, fill the "jobProfileName" & "jobProfileId" to the related jobExecurion.
2.During fetching JobExecution, get related job profile and populate relevant jobProfilename;

Worth thinking about where it is better to populate jobProfileInfo to the JobExecution - as in the PR or straight in the controller, before call manager.

## Pre-Merge Checklist


src/main/java/org/folio/service/manager/input/InputDataManagerImpl.java
```diff
@@ -86,9 +87,10 @@ public void proceed(JsonObject payload, ExportResult exportResult) {
     }, this::handleExportResult);
   }
 
-  protected void initBlocking(JsonObject request, JsonObject requestFileDefinitionJson, JsonObject mappingProfileJson, Map<String, String> params) {
+  protected void initBlocking(JsonObject request, JsonObject requestFileDefinitionJson, JsonObject mappingProfileJson, JsonObject jobProfileJson, Map<String, String> params) {
     ExportRequest exportRequest = request.mapTo(ExportRequest.class);
     MappingProfile mappingProfile = mappingProfileJson.mapTo(MappingProfile.class);
+    JobProfile jobProfile = jobProfileJson.mapTo(JobProfile.class);
```
illia-borysenko, line 93: Yes, I think it's better to populate job profile on the controller level as you proposed instead of passing one more additional parameter to InputManager

andrii-novytskyi: updated

src/main/java/org/folio/rest/impl/DataExportImpl.java
```diff
@@ -64,13 +64,18 @@ public void postDataExportExport(ExportRequest entity, Map<String, String> okapi
...
-                inputDataManager.init(JsonObject.mapFrom(entity), JsonObject.mapFrom(requestFileDefinition), JsonObject.mapFrom(mappingProfile), okapiHeaders);
-                succeededFuture()
-                  .map(PostDataExportExportResponse.respond204())
-                  .map(Response.class::cast)
-                  .onComplete(asyncResultHandler);
-              })
+              .onSuccess(mappingProfile ->
+                jobExecutionService.getById(requestFileDefinition.getJobExecutionId(), tenantId)
+                  .onSuccess(jobExecution ->
+                    jobExecutionService.updateJobExecutionWithJobProfileInfo(jobExecution, jobProfile, tenantId)
```
KVupp, line 70: The 2nd option you mentioned in the approach is actually good enough. We need not populate the job profile name and store it in DB while exporting, just the ID is enough,

ramls/schemas/jobExecution.json
```diff
@@ -34,6 +34,10 @@
       "type": "string",
       "$ref": "../../ramls/raml-util/schemas/uuid.schema"
     },
+    "jobProfileName": {
```
KVupp, line 37: make this a read only field that way we can just populate it in the GET call for jobExecution

KVupp: That way the job profile name is never stale

src/main/java/org/folio/service/job/JobExecutionServiceImpl.java
```diff
@@ -54,16 +58,31 @@
...
   public Future<JobExecution> getById(final String jobExecutionId, final String tenantId) {
-    return jobExecutionDao.getById(jobExecutionId, tenantId)
-      .compose(optionalJobExecution -> {
-        if (optionalJobExecution.isPresent()) {
-          return succeededFuture(optionalJobExecution.get());
+    Promise<JobExecution> jobExecutionPromise = Promise.promise();
+    jobExecutionDao.getById(jobExecutionId, tenantId)
+      .onComplete(optionalJobExecution -> {
+        if (optionalJobExecution.succeeded() && optionalJobExecution.result().isPresent()) {
+          JobExecution jobExecution = optionalJobExecution.result().get();
```
KVupp, line 65: we need to do this for get as well, may be get all the job profiles and just populate the job profile name for the jobs

KVupp: we could add cache later for the job profiles, so that DB is not called multiple times to get the names

src/main/java/org/folio/rest/impl/DataExportImpl.java
andrii-novytskyi: I did not update the job execution while calling get job execution by id, to reduce interactions to DB. The only place where we are populating jobProfileName and updating the object in DB it is here, so I think we need to leave it as is here

andrii-novytskyi: I don`t update the jobExecution object in DB while calling get job execution, to reduce interactions to the DB. The only one place where we are set the jobProfileName to the jobExecution and update the object in the DB is during export, so I think we should leave this logic as is.

src/main/java/org/folio/service/job/JobExecutionServiceImpl.java
andrii-novytskyi: It sounds not really good if we are thinking about performance...why we need to do it for get endpoint? If the jobExecution is new, no jobProfileName there, if it is used for export - jobProfileName will be present in job execution. Maybe we can populate "default" jobProfileName for those jobExecutions, that don`t have this field and when this jobExecution will be used for export, the valid jobProfileName will be populated for it. What do you think?

src/main/java/org/folio/rest/impl/DataExportImpl.java
andrii-novytskyi: why can`t we leave it as is? if we will remove set jobProfileName here, all jobExecutions in the DB will not have this field

src/main/java/org/folio/rest/impl/DataExportImpl.java
```diff
@@ -64,13 +64,18 @@ public void postDataExportExport(ExportRequest entity, Map<String, String> okapi
...
             mappingProfileService.getById(jobProfile.getMappingProfileId(), tenantId)
-              .onSuccess(mappingProfile -> {
-                inputDataManager.init(JsonObject.mapFrom(entity), JsonObject.mapFrom(requestFileDefinition), JsonObject.mapFrom(mappingProfile), okapiHeaders);
-                succeededFuture()
-                  .map(PostDataExportExportResponse.respond204())
-                  .map(Response.class::cast)
-                  .onComplete(asyncResultHandler);
-              })
+              .onSuccess(mappingProfile ->
+                jobExecutionService.getById(requestFileDefinition.getJobExecutionId(), tenantId)
```
KVupp, line 68: we need to refactor this to a helper class or something next time we push anything, getting hard to read. 
we can leave it as is for now

src/main/java/org/folio/service/job/JobExecutionServiceImpl.java
```diff
@@ -33,13 +34,40 @@
...
+                .stream()
+                .filter(jobProfile -> jobProfile.getId().equals(jobExecution.getJobProfileId()))
+                .findFirst()
+                .ifPresent(jobProfile -> jobExecution.setJobProfileName(jobProfile.getName()))))
+            .onFailure(async -> {
+              LOGGER.error("Failed to get jobProfiles from DB while get Job Executions by query for tenant {}, the default name will be used for those jobExecutions, that don`t have jobProfileName", tenantId);
+              jobExecutionCollection.getJobExecutions()
+                .stream()
+                .filter(jobExecution -> StringUtils.isEmpty(jobExecution.getJobProfileName()))
+                .forEach(jobExecution -> jobExecution.setJobProfileName(DEFAULT));
```
KVupp, line 63: I doubt showing inaccurate information is good here, may be we send it without any name and UI can show blank

src/main/java/org/folio/service/job/JobExecutionServiceImpl.java
```diff
@@ -54,16 +82,33 @@
...
+    jobExecutionDao.getById(jobExecutionId, tenantId)
+      .onComplete(optionalJobExecution -> {
+        if (optionalJobExecution.succeeded() && optionalJobExecution.result().isPresent()) {
+          JobExecution jobExecution = optionalJobExecution.result().get();
+          if (StringUtils.isNotEmpty(jobExecution.getJobProfileId())) {
+            jobProfileService.getById(jobExecution.getJobProfileId(), tenantId)
+              .onSuccess(jobProfile -> jobExecution.setJobProfileName(jobProfile.getName())).onFailure(ar -> {
+              if (StringUtils.isEmpty(jobExecution.getJobProfileName())) {
+                LOGGER.error("Failed to get Job Profile with id {} while querying Job Execution with id {}, the default name will be used", jobExecution.getJobProfileId(), jobExecution.getId());
+                jobExecution.setJobProfileName(DEFAULT);
```
KVupp, line 95: same thing here, showing no information is better than incorrect information

src/main/resources/templates/db_scripts/migration_scripts/add_job_profile_name.sql
```diff
@@ -0,0 +1,3 @@
+UPDATE ${myuniversity}_${mymodule}.job_executions
```
KVupp, line 1: you don't need this anymore right?

src/main/java/org/folio/service/job/JobExecutionServiceImpl.java
andrii-novytskyi: they asked for us that name always should be present. Let`s keep it as it is now, I will ask UI guys tomorrow to provide such check on UI, and if they will, I will change this logic later, in the scope of another ticket

KVupp: Thanks!  let's keep that in mind and make sure to verify with Magda as well.

src/main/resources/templates/db_scripts/migration_scripts/add_job_profile_name.sql
KVupp: but you need the job profile Id to be populated ,

andrii-novytskyi: updated

src/main/java/org/folio/service/job/JobExecutionServiceImpl.java
andrii-novytskyi: have a response from Magda that empty string if more preferable. Will tell the UI guys about it

PR #107, opened 2020-06-23
MDEXP-196 selecting fields from the inventory instance in the mapping profile

illia-borysenko:
## Purpose
This PR contains a technical solution how to applying selecting fields for instances with and without underlying SRS records. Most of the changes relate to RuleFactory, MappingService and SrsRecordConvertorService. I also amended rulesDefault.json file and added "id" field values for each rule to have the possibility to select the default rules.


KVupp:
Closing this PR as the code related to this has already been moved as part of various stories

PR #109, opened 2020-06-24
MDEXP-166 - Applied performance improvements

Igor-Gorchakov:
MDEXP-166

## Purpose
The purpose of this PR is to improve timing results for the export process.

## Approach
* Improved efficiency of MappingService:
run the mapping process in parallel threads using parallel stream with pool size = 4, that improves timing on 30%
https://docs.google.com/spreadsheets/d/1HQLZF00Mb7WgEGg4-CnACgH6lV_Rpjg9/edit#gid=1857678284

* Enabled ability to process 2 jobs in parallel:
Increase pool size in InputDataManager from 1 to 2
Increate pool size in ExportManager from 1 to 2

* I do not update 'chunkSize' intentionally and just left it 50. That's because an issue with the 'progress' is tracked. However, the 'chunkSize' definitely affects the timing

* Created an additional Jira to play with 'chunkSize', apply PoolingHttpConnectionManager to reuse HttpConnections, play with SRS_LOAD_PARTITION_SIZE and INVENTORY_LOAD_PARTITION_SIZE
https://issues.folio.org/browse/MDEXP-207

My research about calling the parallel stream from Vertx WorkerExecutor :
* _Calling parallelStream directly from WorkerExecutor does not gain the performance. The ForkJoinPool arranges to execute the ForkJoinTask in the pool the current task is running in, if applicable. So if to call parallelStream in 
 MappingService, the ForkJoinPool pushes tasks to the WorkerExecutor instead of common ForkJoinPool, that's why there is no great perf gain.
There actually is a trick how to execute a parallel operation in a specific pool. If you execute it as a task in ServiceExecutor, it stays there and does not use the pool from WorkerExecutor._

## Pre-Merge Checklist


src/main/java/org/folio/service/mapping/MappingServiceImpl.java
```diff
@@ -39,27 +46,47 @@
...
+    List<String> records = null;
+    try {
+      records = mappingThreadPool.submit(() -> instances.parallelStream()
+        .map(instance -> mapInstance(instance, referenceData, rules))
+        .filter(Optional::isPresent)
+        .map(Optional::get)
+        .collect(Collectors.toList()))
+        .get();
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
```
KVupp, line 72: Do you think it might be worthwhile to add the instances in a debug log that failed to be mapped?

src/main/java/org/folio/service/mapping/MappingServiceImpl.java
```diff
@@ -29,6 +34,8 @@
 @Service
 public class MappingServiceImpl implements MappingService {
   private static final Logger LOGGER = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+  private static final int MAPPING_POOL_SIZE = 4;
```
KVupp, line 37: Is there a reason you choose 4? If I understand NewWork StealingPool by default uses the number of available processors to the JVM

KVupp: Also given that this module can be run in multiple environments like container bare metals, selecting such number would have different effects based on where it is deployed

src/main/java/org/folio/service/mapping/MappingServiceImpl.java
Igor-Gorchakov: Debug log to track failed instances to be mapped, is already in place, line 87

src/main/java/org/folio/service/mapping/MappingServiceImpl.java
Igor-Gorchakov: The reason to use exactly 4 threads is that it is enough 4 threads to noticeably increase the productivity of the MappingService and not enough to load-up all the available processors. 4 threads is a trade-off that leaves CPU resources to handle other operations. The more threads in a pool the more consuming time to manage these threads. It tested Mapping Service using 16 threads and it works worse than using 4.

PR #127, opened 2020-07-21
MDEXP-204: SPIKE: Adding timer job to expire entries

KVupp:
## Purpose
https://issues.folio.org/browse/MDEXP-204

## Approach
There are 2 approaches that have come out from the spike, the details are added to the JIRA ticket above. 
This PR contains a skeleton code for the 2nd approach, of setting up an API with timer interface
The API can also be called manually to expire the entries
To expire the entries the job is set to "FAIL" status
The cleaning up of any remanant files is taken care of by the cleanup job

## Learning
- https://github.com/folio-org/okapi/blob/master/doc/guide.md#timer-interface

### DIFF
--- a/src/main/java/org/folio/dao/JobExecutionDao.java
+++ b/src/main/java/org/folio/dao/JobExecutionDao.java
@@ -1,9 +1,10 @@
 package org.folio.dao;
 
 import io.vertx.core.Future;
-
+import java.util.Date;
+import java.util.List;
 import java.util.Optional;
-
+import org.folio.rest.jaxrs.model.FileDefinition;
 import org.folio.rest.jaxrs.model.JobExecution;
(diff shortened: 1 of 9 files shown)


src/main/java/org/folio/dao/JobExecutionDao.java
```diff
@@ -48,4 +49,6 @@
    * @return future
    */
   Future<Optional<JobExecution>> getById(String jobExecutionId, String tenantId);
+
+  Future<List<JobExecution>> getExpiredEntries(Date lastUpdateDate, String tenantId);
```
andrii-novytskyi, line 53: Please, add a description to this method

KVupp:
Closing this PR as the implementation is being moved to be handled under https://issues.folio.org/browse/MDEXP-246

PR #133, opened 2020-07-29
MDEXP-150 Mapping profile - transformations - retrieve field names

andrii-novytskyi:
## Purpose
https://issues.folio.org/browse/MDEXP-150

## Approach
Provided an endpoint for retrieving transformation fields and mechanism for building transformation field objects.


illia-borysenko:
Approved, but more unit tests should be commited

ramls/dataExport-fieldNames.raml
```diff
@@ -0,0 +1,29 @@
+#%RAML 1.0
+
+title: Data export
+version: v2.0
+protocols: [ HTTP, HTTPS ]
+baseUri: https://github.com/folio-org/mod-data-export
+
+documentation:
+  - title: Data export fiel names API
```
KVupp, line 9: looks like a typo

descriptors/ModuleDescriptor-template.json
```diff
@@ -257,6 +257,18 @@
           ],
           "modulePermissions": [
           ]
+        },
+        {
+          "methods": [
+            "GET"
+          ],
+          "pathPattern": "/data-export/fieldNames",
```
KVupp, line 265: I think we give a better name that is more readable, like transformations or mappingTransformations

andrii-novytskyi: I can rename it here, but in the code, we have transformations already...so that`s why I choose field names

ramls/schemas/fieldName.json
```diff
@@ -0,0 +1,39 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "description": "Field names",
+  "type": "object",
+  "additionalProperties": false,
+    "properties": {
+      "id": {
```
KVupp, line 7: This is not a UUId according to the example above. Also let's not use id as the field name if its not UUID to avoid confusions like this

ramls/dataExport-fieldNames.raml
andrii-novytskyi: ops, thanks!

ramls/schemas/fieldName.json
andrii-novytskyi: I will rename it to the ruleId

src/main/java/org/folio/service/fieldname/FieldNamesService.java
```diff
@@ -0,0 +1,18 @@
...
+
+import io.vertx.core.Future;
+import org.folio.rest.jaxrs.model.FieldName;
+import org.folio.rest.jaxrs.model.FieldNameCollection;
+import org.folio.util.OkapiConnectionParams;
+
+public interface FieldNamesService {
+
+  /**
+   * Searches for {@link FieldNameCollection} by id
```
KVupp, line 11: description needs to change

src/main/java/org/folio/service/fieldname/FiledNamesServiceImpl.java
```diff
@@ -0,0 +1,111 @@
...
+import static org.folio.rest.jaxrs.model.FieldName.RecordType.ITEM;
+import static org.folio.service.fieldname.FieldNameConfig.FIXED_LENGTH_DATA_ELEMENT;
+import static org.folio.service.fieldname.FieldNameConfig.HR_ID;
+import static org.folio.service.fieldname.FieldNameConfig.IDENTIFIERS;
+
+@Service
+public class FiledNamesServiceImpl implements FieldNamesService {
+
+  public static final String REFERENCE_DATA_NAME_KEY = "name";
+  private static final Set<FieldNameConfig> INSTANCE_FIELD_NAME_CONFIGS = EnumSet.of(
```
KVupp, line 34: We better move these constants to a separate file. If it's too much in this PR let's make to do it in the next PRs

src/main/java/org/folio/service/fieldname/FieldNameConfig.java
```diff
@@ -0,0 +1,59 @@
...
+
+import java.util.Map;
+
+public enum FieldNameConfig {
+
+  IDENTIFIERS("identifiers", "$.{recordType}[*].identifiers[?(@identifierTypeId=={id})].value", new IdentifierTypesLoader()),
+  FIXED_LENGTH_DATA_ELEMENT("metadata.createdDate", "$.{recordType}.metadata.createdDate", MetadataParametersConstants.getFixedLengthDataElement()),
+  HR_ID("hrid", "$.{recordType}.hrid");
+
+  private final String id;
```
KVupp, line 14: let's call it fieldId or something let's reserved id only for referring to UUIDs

andrii-novytskyi: sure, thanks

src/main/java/org/folio/service/fieldname/loader/IdentifierTypesLoader.java
```diff
@@ -0,0 +1,22 @@
+package org.folio.service.fieldname.loader;
+
+import io.vertx.core.json.JsonObject;
+import org.folio.clients.InventoryClient;
+import org.folio.util.OkapiConnectionParams;
+
+import java.util.Map;
+
+public class IdentifierTypesLoader implements ReferenceDataLoader {
```
KVupp, line 9: Why can't we reuse the ReferenceDataProvider? It already loads all the required reference data anyway

KVupp: We are essentially duplicating the same calls here , and will end up creating 10 more classes for the same

ramls/samples/fieldName.sample
```diff
@@ -0,0 +1,7 @@
+{
```
KVupp, line 1: Did you guys talk this structure through with UI guys, this is different from the transformations that are required by the mapping profile API. 
So when retrieving back they should be able to map these 2

ramls/samples/fieldName.sample
```diff
@@ -0,0 +1,7 @@
...
+   "id":"instance.identifiers.invalid.issn",
+   "displayNameKey":"instance.identifiers",
+   "referenceDataValue":"Invalid ISSN",
```
KVupp, line 4: I thought we are going to use a dot formatted name here how will the UI translate this value?

src/main/java/org/folio/service/fieldname/loader/IdentifierTypesLoader.java
andrii-novytskyi: I discussed it with Illia, and we thought to create a ticket for adding cach for this and do this work in the scope of that ticket

ramls/samples/fieldName.sample
andrii-novytskyi: As far as I know, `referenceDataValue` is untranslatable. Yes, I showed the structure of response to Victor
I will clarify is it translatable with Magda

src/main/java/org/folio/service/fieldname/FieldNameConfig.java
```diff
@@ -0,0 +1,59 @@
...
+
+import org.folio.service.fieldname.loader.IdentifierTypesLoader;
+import org.folio.service.fieldname.loader.ReferenceDataLoader;
+
+import java.util.Map;
+
+public enum FieldNameConfig {
+
+  IDENTIFIERS("identifiers", "$.{recordType}[*].identifiers[?(@identifierTypeId=={id})].value", new IdentifierTypesLoader()),
+  FIXED_LENGTH_DATA_ELEMENT("metadata.createdDate", "$.{recordType}.metadata.createdDate", MetadataParametersConstants.getFixedLengthDataElement()),
```
KVupp, line 11: I am guessing this part is incorrect, why do we need languages for metadata. We need to correct this

src/main/java/org/folio/service/fieldname/loader/IdentifierTypesLoader.java
KVupp: Right even adding cache is a duplicate right, that class is already doing what we need

KVupp: Is there a special reason you need  separate classes?

src/test/java/org/folio/service/fieldname/FieldNameServiceUnitTest.java
```diff
@@ -0,0 +1,79 @@
...
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.doCallRealMethod;
+
+@RunWith(VertxUnitRunner.class)
+@ExtendWith(VertxExtension.class)
+@ExtendWith(MockitoExtension.class)
+class FieldNameServiceUnitTest extends RestVerticleTestBase {
```
KVupp, line 36: If this is a unit test why does it need to extend this class?

src/main/java/org/folio/service/fieldname/FieldNameConfig.java
andrii-novytskyi: I will remove FIXED_LENGTH_DATA_ELEMENT field for now, but will not remove the mechanism of adding metadata... We can easily do it in the future if we will not need it

src/main/java/org/folio/service/fieldname/loader/IdentifierTypesLoader.java
andrii-novytskyi: I will change it

ramls/samples/tansformationField.sample
```diff
@@ -0,0 +1,7 @@
+{
+   "id":"instance.identifiers.invalid.issn",
```
KVupp, line 2: pls change the field name to ruleId here as well

ramls/samples/transformationFieldCollection.sample
```diff
@@ -0,0 +1,11 @@
+{
+   "transformationFields":[
+      {
+         "id":"instance.identifiers.invalid.issn",
```
KVupp, line 4: ruleId

src/main/java/org/folio/service/fieldname/FieldNameConfig.java
KVupp: yes we will need the metadata mechanism. We shouldn't be removing that

ramls/samples/tansformationField.sample
```diff
@@ -0,0 +1,7 @@
+{
```
KVupp, line 1: This filename has a typo. Please correct if before merge

ramls/samples/fieldName.sample
andrii-novytskyi: discussed with Victor, all ok, going to merge

PR #139, opened 2020-08-05
MDEXP-217 Provide transformation fields with reference data

andrii-novytskyi:
https://issues.folio.org/browse/MDEXP-217

## Purpose
The purpose is to add fields that specified in the ticket to the response of /transformationFields.

## Approach
Added the fields with reference data to the enum set
Changed JsonPathBuilder, provide a way to path the actual value of the reference data without id.

### DIFF
--- a/src/main/java/org/folio/service/transformationfields/TransformationFieldsConfig.java
+++ b/src/main/java/org/folio/service/transformationfields/TransformationFieldsConfig.java
@@ -14,23 +14,43 @@ import static org.folio.service.mapping.referencedata.ReferenceDataImpl.MODES_OF
  */
 public enum TransformationFieldsConfig {
 
-  ALTERNATIVE_TITLES("alternativeTitles", "$.{recordType}[*].alternativeTitles[?(@.alternativeTitle=={value})]", ALTERNATIVE_TITLE_TYPES),
-  HR_ID("hrid", "$.{recordType}.hrid"),
+  //Common Fields
   ID("id", "$.{recordType}.id"),
+  HR_ID("hrid", "$.{recordType}.hrid"),
   IDENTIFIERS("identifiers", "$.{recordType}[*].identifiers[?(@.identifierTypeId=={id})].value", IDENTIFIER_TYPES),
-  INSTANCE_ID("instanceId", "$.{recordType}.instanceId"),
   INSTANCE_TYPE("instanceTypeId", "$.{recordType}.instanceTypeId", INSTANCE_TYPES),
 
   MATERIAL_TYPE_ID("materialTypeId", "$.{recordType}[*].materialTypeId", MATERIAL_TYPES),
   MODE_OF_ISSUANCE_ID("modeOfIssuanceId", "$.{recordType}.modeOfIssuanceId", MODES_OF_ISSUANCE),
 
+  SOURCE("source", "$.{recordType}.source"),
   METADATA_CREATED_DATE("metadata.createdDate", "$.{recordType}.metadata.createdDate", MetadataParametersConstants.getFixedLengthDataElement()),
-  METADATA_CREATED_BY_USER_ID("metadata.createdByUserId", "$.{recordType}.metadata.createdByUserId"),
   METADATA_UPDATED_DATE("metadata.updatedDate", "$.{recordType}.metadata.updatedDate"),
+  METADATA_CREATED_BY_USER_ID("metadata.createdByUserId", "$.{recordType}.metadata.createdByUserId"),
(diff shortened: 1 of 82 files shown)


src/main/java/org/folio/service/transformationfields/TransformationFieldsConfig.java
```diff
@@ -2,25 +2,35 @@
...
-  IDENTIFIERS("identifiers", "$.{recordType}[*].identifiers[?(@identifierTypeId=={id})].value", IDENTIFIER_TYPES),
+
   SOURCE("source", "$.{recordType}.source"),
   METADATA_CREATED_DATE("metadata.createdDate", "$.{recordType}.metadata.createdDate", MetadataParametersConstants.getFixedLengthDataElement()),
   METADATA_UPDATED_DATE("metadata.updatedDate", "$.{recordType}.metadata.updatedDate"),
   METADATA_CREATED_BY_USER_ID("metadata.createdByUserId", "$.{recordType}.metadata.createdByUserId"),
   METADATA_UPDATED_BY_USER_ID("metadata.updatedByUserId", "$.{recordType}.metadata.updatedByUserId"),
 
   //Instance specific fields
+  ALTERNATIVE_TITLES("alternativeTitles", "$.instance[*].alternativeTitles[?(@.alternativeTitle=={value})]", ALTERNATIVE_TITLE_TYPES),
```
KVupp, line 28: I know you showed me the example yesterday Andrii, but comparing alternative tile is not the right way, atleast from UI the typeId is mandatory. If that is not present we can ignore that mapping, but this must be aletrnativeTitleTypeId

KVupp: because the title is record specific it will never match anything

andrii-novytskyi: ok, updated

src/test/resources/mapping/expectedTransformationFields.json
```diff
@@ -46,6 +46,20 @@
 			"path": "$.holdings.metadata.updatedDate",
 			"recordType": "HOLDINGS"
 		},
+    {
+      "fieldId":"instance.alternativetitles.cover.title",
+      "displayNameKey":"instance.alternativeTitles",
+      "referenceDataValue":"Cover title",
+      "path":"$.instance[*].alternativeTitles[?(@.alternativeTitle==Cover title)]",
```
KVupp: yea I think this will not work because , I can give any title I like
example:
` "alternativeTitles" : [ {
      "alternativeTitleTypeId" : "5c364ce4-c8fd-4891-a28d-bb91c9bcdbfb",
      "alternativeTitle" : "kv"
    }, {
      "alternativeTitleTypeId" : "09964ad1-7aed-49b8-8223-a4c105e3ef87",
      "alternativeTitle" : "testing alt"
    } ],`

src/test/resources/mapping/expectedTransformationFields.json
```diff
@@ -68,44 +82,58 @@
...
 			"fieldId": "instance.identifiers.system.control.number",
 			"displayNameKey": "instance.identifiers",
 			"referenceDataValue": "System control number",
-			"path": "$.instance[*].identifiers[?(@identifierTypeId==7e591197-f335-4afb-bc6d-a6d76ca3bace)].value",
+			"path": "$.instance[*].identifiers[?(@.identifierTypeId==7e591197-f335-4afb-bc6d-a6d76ca3bace)].value",
 			"recordType": "INSTANCE"
 		},
+    {
+      "fieldId":"instance.instancetypeid.other",
+      "displayNameKey":"instance.instanceTypeId",
```
KVupp, line 125: This will need a translation funtion right, because the Id must be translated to value

src/test/resources/mapping/expectedTransformationFields.json
andrii-novytskyi: agree, thank you, updated the field

src/test/resources/mapping/expectedTransformationFields.json
andrii-novytskyi: yes, this work required in stories about applying. We have a map in the rule factory, there we will add field id as a key, and function name as a value

andrii-novytskyi: added function

KVupp: The other functions are still missing, but we can add it in the implementation

PR #140, opened 2020-08-05
MDEXP-210 and MDEXP-218- Simple fields for Instance and Item

KVupp:
https://issues.folio.org/browse/MDEXP-210
https://issues.folio.org/browse/MDEXP-218

## Purpose
Display transformation fields for Instances and Items. 
This PR deals with just simple fields

## Approach
- Added the simple fields to the enum set
- Added a comparator for sort

## Pre-Merge Checklist

### DIFF
--- a/src/main/java/org/folio/service/transformationfields/TransformationFieldsConfig.java
+++ b/src/main/java/org/folio/service/transformationfields/TransformationFieldsConfig.java
@@ -8,21 +8,36 @@ import static org.folio.service.mapping.referencedata.ReferenceDataImpl.IDENTIFI
  * Initial data for the transformation field. While extending the enum, put new values in alphabetical order
  */
 public enum TransformationFieldsConfig {
-
+  //Common Fields
   ID("id", "$.{recordType}.id"),
   HR_ID("hrid", "$.{recordType}.hrid"),
-  INSTANCE_ID("instanceId", "$.holdings.instanceId"),
   IDENTIFIERS("identifiers", "$.{recordType}[*].identifiers[?(@identifierTypeId=={id})].value", IDENTIFIER_TYPES),
   SOURCE("source", "$.{recordType}.source"),
-  EDITIONS("editions", "$.{recordType}.editions"),
-  SUBJECTS("subjects", "$.{recordType}.subjects"),
-  LANGUAGES("languages", "$.{recordType}.languages"),
-  TITLE("title", "$.{recordType}.title"),
-
   METADATA_CREATED_DATE("metadata.createdDate", "$.{recordType}.metadata.createdDate", MetadataParametersConstants.getFixedLengthDataElement()),
   METADATA_UPDATED_DATE("metadata.updatedDate", "$.{recordType}.metadata.updatedDate"),
   METADATA_CREATED_BY_USER_ID("metadata.createdByUserId", "$.{recordType}.metadata.createdByUserId"),
-  METADATA_UPDATED_BY_USER_ID("metadata.updatedByUserId", "$.{recordType}.metadata.updatedByUserId");
+  METADATA_UPDATED_BY_USER_ID("metadata.updatedByUserId", "$.{recordType}.metadata.updatedByUserId"),
+
+  //Instance specific fields
+  EDITIONS("editions", "$.instance.editions"),
+  LANGUAGES("languages", "$.instance.languages"),
+  SUBJECTS("subjects", "$.instance.subjects"),
+  TITLE("title", "$.instance.title"),
+
+  //Holdings specific Fields
+  INSTANCE_ID("instanceId", "$.holdings.instanceId"),
+
+  //Item specific fields
+  BARCODE("barcode", "$.item.barcode"),
+  CHRONOLOGY("chronology", "$.item.chronology"),
+  COPYNUMBER("copyNumber", "$.item.copyNumber"),
+  DESCRIPTION_OF_PIECES("descriptionOfPieces", "$.item.descriptionOfPieces"),
+  ENUMERATION("enumeration", "$.item.enumeration"),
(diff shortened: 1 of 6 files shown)


src/test/java/org/folio/service/transformationfields/TransformationFieldsServiceUnitTest.java
```diff
@@ -96,6 +96,8 @@ void getFieldNamesShouldReturnValidFields(VertxTestContext context) {
         TransformationFieldCollection transformationFieldCollection = ar.result();
         transformationFieldCollection.getTransformationFields()
           .forEach(transformationField -> checkIfActualFieldEqualToExpected(expectedFields.get(transformationField.getFieldId()), transformationField));
+
+       System.out.println(transformationFieldCollection);
```
andrii-novytskyi, line 100: I think this sout shout be removed)

src/main/java/org/folio/service/transformationfields/TransformationFieldsConfig.java
```diff
@@ -8,16 +8,36 @@
...
-
+  SOURCE("source", "$.{recordType}.source"),
   METADATA_CREATED_DATE("metadata.createdDate", "$.{recordType}.metadata.createdDate", MetadataParametersConstants.getFixedLengthDataElement()),
   METADATA_UPDATED_DATE("metadata.updatedDate", "$.{recordType}.metadata.updatedDate"),
   METADATA_CREATED_BY_USER_ID("metadata.createdByUserId", "$.{recordType}.metadata.createdByUserId"),
-  METADATA_UPDATED_BY_USER_ID("metadata.updatedByUserId", "$.{recordType}.metadata.updatedByUserId");
+  METADATA_UPDATED_BY_USER_ID("metadata.updatedByUserId", "$.{recordType}.metadata.updatedByUserId"),
+
+  //Instance specific fields
+  EDITIONS("editions", "$.instance.editions"),
```
andrii-novytskyi, line 22: I understand that is specific for this type of record, but maybe it is better to use "$.{recordType}.editions" for jsonPath,

KVupp: Is there a reason? I left the one's that will be common among multiple types with recordType.  
The advantage in this way we will be able to easily keep track of which fields are required for each Record type

src/test/java/org/folio/service/transformationfields/TransformationFieldsServiceUnitTest.java
KVupp: yup removed

src/main/java/org/folio/service/transformationfields/TransformationFieldsConfig.java
andrii-novytskyi: ok, it is not a big deal

src/test/java/org/folio/service/transformationfields/TransformationFieldsServiceUnitTest.java
andrii-novytskyi: I do not see, please commit changes

src/test/resources/mapping/expectedTransformationFields.json
```diff
@@ -1,231 +1,257 @@
 {
-  "transformationFields": [
```
andrii-novytskyi, line 2: In the PR for MDEXP-217, I added tests  that check the equals of expected and actual response, to check the order of the fields,, but let`s leave it as it is here, I will modify the file again and commit in my PR

src/test/java/org/folio/service/transformationfields/TransformationFieldsServiceUnitTest.java
KVupp: I think you may have to refresh, here's the commit
https://github.com/folio-org/mod-data-export/pull/140/commits/bebd0301bf9c151a356c8352ab8414421bc80b50

src/main/java/org/folio/service/transformationfields/TransformationFieldsServiceImpl.java
```diff
@@ -46,6 +42,7 @@
     transformationFields.addAll(buildTransformationFields(INSTANCE, TransformationConfigConstants.INSTANCE_FIELDS_CONFIGS, referenceData));
     transformationFields.addAll(buildTransformationFields(HOLDINGS, TransformationConfigConstants.HOLDINGS_FIELDS_CONFIGS, referenceData));
     transformationFields.addAll(buildTransformationFields(ITEM, TransformationConfigConstants.ITEM_FIELDS_CONFIGS, referenceData));
+    Collections.sort(transformationFields, (field1,field2) -> field1.getFieldId().compareTo(field2.getFieldId()));
```
andrii-novytskyi, line 45: Not sure if it works the way we need it. I think the instance fields should be first, please ask Magda about that, because as I understand with this comparator, the holding fields are first

KVupp: Correct, I already confirmed with Magda, she agreed to leave the order this way for now
[3:14 PM] Magda Zacharska
    Let's leave this order for now. Once we hear from them that they want instances first followed by holdings and items  then we will adjust.

andrii-novytskyi: ok, then I will remove those tests for equals responses from my PR.
Also, I think it is better to use List.sort instead of Collection.sort, we can write it as: transformationFields.sort(Comparator.comparing(TransformationField::getFieldId));

KVupp: Probably there is not much difference. Given that we will revisit this logic anyway once we get proper requirements, this will probably go away

PR #144, opened 2020-08-11
MDEXP-248 - Mapping profile - provide field name -  item - effective call number

Igor-Gorchakov:
https://issues.folio.org/browse/MDEXP-248

## Purpose
Provide the effective call number field name so that UI can display it correctly.

Purpose/Overview:
Provide the effective call number field name so that UI can display it correctly.

Requirements/Scope:
The current implementation of effectiveCallNumberComponent consists of multiple elements: Each element of the component should be displayed in the UI in separate fields:

Item - Effective call number
Item - Effective call number - Prefix
Item - Effective call number - Suffix
Item - Effective call number type
Created transformation label:

Holdings - Call number - Call number prefix
Holdings - Call number - Call number
Holdings - Call number - Call number suffix
Holdings - Call number - Call number type

## Pre-Merge Checklist


src/test/resources/mapping/expectedTransformationFields.json
```diff
@@ -321,7 +321,39 @@
...
+    },
+    {
+      "fieldId": "item.callnumbertype.dewey.decimal.classification",
+      "displayNameKey": "item.callNumberType",
+      "referenceDataValue": "Dewey Decimal classification",
+      "path": "$.item.effectiveCallNumberComponents.typeId",
+      "recordType": "ITEM"
+    },
+    {
+      "fieldId": "item.callnumbertype.lc.modified",
```
KVupp, line 351: This data means that we are allowing for each call number type to go to a different field, Should it be that way? or is it just that we need to map the entered call number type id to value. Both are quite different. Can you confirm that once

src/main/java/org/folio/service/mapping/referencedata/ReferenceDataProvider.java
```diff
@@ -53,6 +54,7 @@ public ReferenceData getReferenceDataForTransformationFields(OkapiConnectionPara
     referenceData.put(LOAN_TYPES, inventoryClient.getLoanTypes(okapiConnectionParams));
     referenceData.put(MATERIAL_TYPES, inventoryClient.getMaterialTypes(okapiConnectionParams));
     referenceData.put(MODES_OF_ISSUANCE, inventoryClient.getModesOfIssuance(okapiConnectionParams));
+    referenceData.put(CALLNUMBER_TYPES, inventoryClient.getCallNumberTypes(okapiConnectionParams));
```
andrii-novytskyi, line 57: I think the name should be CALL_NUMBER_TYPES

src/test/java/org/folio/rest/impl/MockServer.java
```diff
@@ -366,6 +368,17 @@ private void handleGetLoanTypes(RoutingContext ctx) {
     }
   }
 
+  private void handleGetCallNumberTypes(RoutingContext ctx) {
+    logger.info("handle Get call number types: ", ctx.request().path());
+    try {
+      JsonObject loanTypes = new JsonObject(RestVerticleTestBase.getMockData(CALLNUMBER_TYPES_MOCK_DATA_PATH));
```
andrii-novytskyi, line 374: "callNumberTypes"

src/test/resources/mapping/expectedTransformationFields.json
Igor-Gorchakov: I think it should work like instance.identifiers, @andrii-novytskyi  what do you think ?

Igor-Gorchakov: Confirmed with Magda, it should not work like instance.identifiers, no need to use reference data

src/test/java/org/folio/rest/impl/MockServer.java
```diff
@@ -138,7 +139,6 @@ private Router defineRoutes() {
     router.get(resourcesPath(HOLDING)).handler(ctx -> handleGetHoldingRecord(ctx));
     router.get(resourcesPath(ITEM)).handler(ctx -> handleGetItemRecord(ctx));
     router.get(resourcesPath(CONFIGURATIONS)).handler(ctx -> handleGetConfigurations(ctx));
-
```
andrii-novytskyi, line 141: I don`t see the call number response from MockServer

src/test/java/org/folio/util/ReferenceDataResponseUtil.java
```diff
@@ -142,5 +142,4 @@
     });
     return stringJsonObjectMap;
   }
-
```
andrii-novytskyi, line 145: why you removed call number response here?

andrii-novytskyi:
please, add call number permission to the module descriptor

andrii-novytskyi:
Please, add  test about call number to InventoryClientTest

src/main/java/org/folio/service/transformationfields/TransformationFieldsConfig.java
```diff
@@ -60,7 +60,11 @@
   PERMANENT_LOAN_TYPE_ID("permanentLoanTypeId", "$.item.permanentLoanTypeId", LOAN_TYPES),
   STATUS("status", "$.item.status.name"),
   VOLUME("volume", "$.item.volume"),
-  YEARCAPTION("yearCaption", "$.item.yearCaption");
+  YEARCAPTION("yearCaption", "$.item.yearCaption"),
+  CALL_NUMBER("callNumber", "$.item.effectiveCallNumberComponents.callNumber"),
+  CALL_NUMBER_PREFIX("callNumberPrefix", "$.item.effectiveCallNumberComponents.prefix"),
+  CALL_NUMBER_SUFFIX("callNumberSuffix", "$.item.effectiveCallNumberComponents.suffix"),
+  CALL_NUMBER_TYPE("callNumberType", "$.item.effectiveCallNumberComponents.typeId");
```
KVupp, line 67: This needs metadata to be sent right? Because we have to replace the UUID with corresponding value?

illia-borysenko: As I understood from Magda's answer in jira call number type should be set as id to marc file. In this case, we don't need metadata and jsonPath is enough

PR #150, opened 2020-08-25
MDEXP-251 Mapping profile - provide field name - locations - holdings…

illia-borysenko:
MDEXP-251 - Mapping profile - provide field name - locations - holdings and items

- Add new location fields and poc of the solution for building translation function with necessary parameters in RuleFactory

-  Add fix for an alternative title field


src/main/java/org/folio/service/mapping/RuleFactory.java
```diff
@@ -80,6 +84,9 @@
   private static final String SET_METADATA_UPDATED_DATE_FUNCTION = "set_transaction_datetime";
   private static final String SET_METADATA_CREATED_DATE_FUNCTION = "set_fixed_length_data_elements";
   private static final String MOD_OF_ISSUANCE_ID_FUNCTION = "set_mode_of_issuance_id";
+  private static final String SET_LOCATIONS_FUNCTION_NEW = "set_locations_function_new"; // a new function for locations, should replace old one
```
KVupp, line 87: If we are replacing old one, we'll need a migrations cript

src/main/java/org/folio/service/mapping/RuleFactory.java
```diff
@@ -91,6 +98,9 @@
     .put(SET_METADATA_CREATED_DATE_FIELD_ID, SET_METADATA_CREATED_DATE_FUNCTION)
     .put(CALL_NUMBER_TYPE_FIELD_ID, SET_CALL_NUMBER_TYPE_ID_FUNCTION) // implement 'set_call_number_type_id'
     .put(MOD_OF_ISSUANCE_ID, MOD_OF_ISSUANCE_ID_FUNCTION)
+    .put(PERMANENT_LOCATION_NAME, SET_LOCATIONS_FUNCTION_NEW)// implement new function for locations
```
andrii-novytskyi, line 101: does this comment mean that we need a PR for generate-marc-utils?

andrii-novytskyi (approved):

src/main/java/org/folio/service/mapping/RuleFactory.java
KVupp: ok, so I think we need keep the function name in  generate-marc-utils and the name here the same. Are you going to modify it?

illia-borysenko: Yeah,  I'm modifying it in scope of MDEXP-252

PR #153, opened 2020-09-01
FOLIO-2706 Use FQDN .dev.folio.org

dcrossleyau:


dcrossleyau:
Tests seems to be failing during mvn phase, for a reason not related to this PR.

When fixed, then please merge this.

PR #159, opened 2020-09-03
MDEXP-237: Omit calling SRS for custom Profile

KVupp:
https://issues.folio.org/browse/MDEXP-237

## Purpose
- For Q3 -2020 for records generated from custom profile we are not fetching records from SRS, and always need to generate on the fly

## Approach
- Change to check if custom mapping profile and call only inventory in that case
- Just disabled the old tests with comments
- Added a new method to generate all the holdings and items json mapping
- Corrected mapping path for call number fields
- Added new tests that test all the fields for holdings

ADDITIONAL
- Removed duplicate values from ReferenceDataImpl. Left material types in there because it uses different key. May need to clean that up later

#### TODOS and Open Questions
- [ ] Possibly test this on rancher if all the holding mappings are being generated correctly

## Pre-Merge Checklist


src/main/java/org/folio/service/transformationfields/TransformationConfigConstants.java
```diff
@@ -45,7 +45,7 @@
     ELECTRONIC_ACCESS_LINKTEXT_DEFAULT,
     ELECTRONIC_ACCESS_MATERIALS_SPECIFIED_DEFAULT,
     ELECTRONIC_ACCESS_PUBLICNOTE_DEFAULT,
-    ID,
+    HOLDINGS_UUID,
```
andrii-novytskyi, line 48: ID is a common field for all types, why we specified it for holdings?

andrii-novytskyi:
Don`t we need to wide MappingServiceUnitTest with all fields for holdings in the scope of this ticket?

illia-borysenko (approved):

KVupp:
> Don`t we need to wide MappingServiceUnitTest with all fields for holdings in the scope of this ticket?

Good point Andrii , I added that test. On the way also changed a few things , added those in the description:)

src/main/java/org/folio/service/transformationfields/TransformationConfigConstants.java
KVupp: yup that was a left over cleanup, changed it now. Thanks !

andrii-novytskyi (approved):

PR #163, opened 2020-09-10
MDEXP-246 Implement expire stuck jobs

andrii-novytskyi:
https://issues.folio.org/browse/MDEXP-246

## Purpose
Implement the spike - https://issues.folio.org/browse/MDEXP-204

## Approach
Implement changes from the spike:
- Create new endpoint;
- Added new field to JobExecution;
- Created migration script

## Pre-Merge Checklist


.rancher-pipeline.yml
```diff
@@ -16,10 +16,13 @@ stages:
     steps:
       - applyAppConfig:
           catalogTemplate: p-rlq5p:concorde-helmcharts-mod-data-export
-          version: 0.1.10
+          version: 0.1.26
           answers:
             image.repository: docker.dev.folio.org/mod-data-export
             image.tag: concorde-latest
+            javaOptions: -XX:MaxRAMPercentage=85.0 -XX:+UseG1GC -Dbucket.name=kruthi-test
```
andrii-novytskyi, line 23: @KVupp Is it ok if I will add this bucket name, or it is better to wait for the proper?

KVupp: The actual buckets have been created I believe Andrii, the story seems to be closed. Let's test that out and change it to proper names

andrii-novytskyi:
Verified migration script on local box:
image

src/main/java/org/folio/dao/impl/JobExecutionDaoImpl.java
```diff
@@ -116,6 +120,28 @@
...
+      LOGGER.error("Error during getting fileDefinition entries by expired date", e);
+      promise.fail(e);
+    }
+    return promise.future().map(Results::getResults);
 
+  }
+
+  private Criterion constructExpiredEntriesCriterion(Date expirationDate) {
+    Criterion criterion = new Criterion();
+    Criteria lastUpdateDateCriteria = new Criteria();
```
KVupp, line 139: Shouldn't we check if the job status is in progress. What if the job already completed?

src/main/java/org/folio/service/job/JobExecutionService.java
```diff
@@ -79,4 +79,13 @@
    * @return future
    */
   Future<JobExecution> incrementCurrentProgress(String jobExecutionId, final int exported, final int failed, String tenantId);
+
+  /**
+   * Update status of expired job executions to fail
+   *
+   * @param tenantId tenant id
+   * @return future with list of expire {@link JobExecution}
```
KVupp, line 87: there is no list here right?

src/main/resources/templates/db_scripts/migration_scripts/add_last_updated_date_to_job.sql
```diff
@@ -0,0 +1,3 @@
+UPDATE diku_mod_data_export.job_executions
+SET jsonb = jsonb_set(jsonb, '{lastUpdatedDate}', to_jsonb(now()), true)
```
KVupp, line 2: I think it's better to update the lastUpdated date to the completedDate, instead of now, as these are existing jobs

src/test/java/org/folio/service/job/JobExecutionServiceUnitTest.java
```diff
@@ -363,6 +369,32 @@ void shouldPrepareJobExecutionSuccessfully_whenJobExecutionStartDateIsNull(Vertx
     });
   }
 
+  @Test
+  void expireJobExecutionsShouldSetStatusToFail(VertxTestContext context) {
+    // given
+    JobExecution jobExecution = new JobExecution()
+      .withExportedFiles(Sets.newHashSet())
```
KVupp, line 376: Can you please write a new test/or modify this test to consider the time factor as well. 
This is to make sure that we are getting back entries for the pas hour correctly

KVupp: will be a new test as it will be integration test.

src/main/java/org/folio/service/job/JobExecutionService.java
andrii-novytskyi: ups, thanks!

src/main/resources/templates/db_scripts/migration_scripts/add_last_updated_date_to_job.sql
andrii-novytskyi: @KVupp I thought about it, but not every job has this field, for example, new doesn`t have it, and if we will have job which is stuck, it will not have this date too. Do you think it is an issue if we will add the current date for all jobs?

illia-borysenko: I agree with Andry completedDate doesn't cover all cases

KVupp: That's a good point , my thought was it might create a confusion as to why the dates are different later if someone is looking at the records. I think we can use the completedDate if present, and if not use current date?

andrii-novytskyi: ok, I will modify the script, verify, and then push changes

src/main/java/org/folio/dao/impl/JobExecutionDaoImpl.java
andrii-novytskyi: sure, will add

KVupp (approved):
Looks good now. Thanks!

PR #175, opened 2020-10-08
MDEXP-294 Export records using CQL

illia-borysenko:
## Description
https://issues.folio.org/browse/MDEXP-294

Implement new data export flow for export records using CQL file. Update file definition schema to add a new field uploadFormat with possible values to use specific export flow base of this field value. Use new endpoint /instance-bulk/ids to retrieve UUIDs by CQL query and use for export.

## Pre-Merge Checklist

  - [x] Were there any schema changes?
  
  - [x] Do they contain the appropriate level of detail?  Which endpoints/schemas changed, etc.
  - [x] Do they have all they appropriate links to blocked/related issues?


KVupp:
Please add a description, please fill the pre merge checklist

ramls/schemas/fileDefinition.json
```diff
@@ -36,6 +36,16 @@
       "description": "Size of the file in Kbyte",
       "type": "integer"
     },
+    "format": {
+      "description": "Format of exported file",
```
KVupp, line 40: What do you mean by format of exported file, the output? If so why would csv be the default? our output default is mrc

illia-borysenko: Ok, it makes sense, I will change it

KVupp: I still don't get it Illia, is it the format of the uploaded file or the file being exported?

andrii-novytskyi: Format of an uploaded file, better to name uploadedFormat or something like this

ramls/schemas/fileDefinition.json
```diff
@@ -36,6 +36,16 @@
...
+      "type": "string",
+      "enum": [
+        "csv",
+        "cql",
+        "mrc"
```
andrii-novytskyi, line 45: I think upload file format can not be mrc

src/main/java/org/folio/service/file/upload/FileUploadService.java
```diff
@@ -36,6 +36,14 @@
    */
   Future<FileDefinition> completeUploading(FileDefinition fileDefinition, String tenantId);
 
+  /**
+   * Saves given file data to the {@link FileStorage}
+   *
+   * @param fileDefinition {@link FileDefinition}
```
andrii-novytskyi, line 42: not all params in doc

src/main/java/org/folio/service/file/upload/FileUploadServiceImpl.java
```diff
@@ -43,14 +54,34 @@ public FileUploadServiceImpl(@Autowired FileStorage fileStorage, @Autowired File
 
   @Override
   public Future<FileDefinition> saveFileChunk(FileDefinition fileDefinition, byte[] data, String tenantId) {
-    return fileStorage.saveFileDataAsync(data, fileDefinition);
+    if (data.length > 0) {
+      return fileStorage.saveFileDataAsync(data, fileDefinition)
+        .compose(ar -> updateFileDefinitionWithJobExeution(new JobExecution(), fileDefinition, tenantId));
```
andrii-novytskyi, line 59: wrong method name

ramls/schemas/fileDefinition.json
illia-borysenko: Guys, I don't think it matters, as for me it's the format of the file content, it can be used as we want, but to be consistent we should set it for all file definitions, and export file definition too. Therefore I added "mrc" type as well

andrii-novytskyi:
@illia-borysenko I think we need to update module descriptor

ramls/schemas/fileDefinition.json
illia-borysenko: commented above

andrii-novytskyi:
Also, if we upload cql file, we don`t need to read file lines to get the total count, because we already have it. Please add a check in SourceReader

ramls/schemas/fileDefinition.json
KVupp: Please change the description, as this is filedefinition, please remove mrc to avoid any confusion, we will never allow for uploading mrc files

src/main/java/org/folio/service/file/upload/FileUploadServiceImpl.java
```diff
@@ -59,4 +90,17 @@ public FileUploadServiceImpl(@Autowired FileStorage fileStorage, @Autowired File
       .compose(fileDefinition -> fileDefinitionService.update(fileDefinition.withStatus(ERROR), tenantId));
   }
 
+
+  private Future<FileDefinition> updateFileDefinitionWithJobExeution(JobExecution jobExecution, FileDefinition fileDefinition, String tenantId) {
+    return jobExecutionService.save(jobExecution, tenantId)
+      .compose(savedJob -> fileDefinitionService.update(fileDefinition.withJobExecutionId(savedJob.getId()), tenantId));
+  }
+
+  private JobExecution getJobExecutionWithProgress(int total) {
```
Igor-Gorchakov, line 99: do we really need this method with only one line of code? is it used twice somewhere?

illia-borysenko: usage of such method make code more readable in the place where it is called

andrii-novytskyi (approved):

src/main/java/org/folio/service/file/upload/FileUploadServiceImpl.java
```diff
@@ -43,14 +56,34 @@ public FileUploadServiceImpl(@Autowired FileStorage fileStorage, @Autowired File
...
+  @Override
+  public Future<FileDefinition> saveUUIDsByCQL(FileDefinition fileDefinition, String query, OkapiConnectionParams params) {
+    Optional<JsonObject> instancesUUIDs = inventoryClient.getInstancesBulkUUIDs(query, params);
+    List<String> ids = new ArrayList<>();
+    if (instancesUUIDs.isPresent()) {
+      JsonArray jsonIds = instancesUUIDs.get().getJsonArray("ids");
+      if (jsonIds.size() > 0) {
+        for (Object id : jsonIds) {
+          ids.add((String) id);
+        }
```
Igor-Gorchakov, line 80: just put it here
 new JobExecution().withProgress(new Progress().withTotal(String.valueOf(total)));

src/main/java/org/folio/clients/InventoryClient.java
```diff
@@ -40,6 +43,8 @@
 import static org.folio.util.ExternalPathResolver.MATERIAL_TYPES;
 import static org.folio.util.ExternalPathResolver.resourcesPathWithPrefix;
 
+import javax.swing.text.html.Option;
```
Igor-Gorchakov, line 46: Why do we need this class from 'javax.swing' package? Can we use something else instead?

src/main/java/org/folio/service/file/reader/LocalStorageCsvSourceReader.java
```diff
@@ -59,7 +60,7 @@ public void close() {
 
   @Override
   public long totalCount() {
-    if(nonNull(fileDefinition)) {
+    if(nonNull(fileDefinition) && !CQL.equals(fileDefinition.getUploadFormat())) {
```
Igor-Gorchakov, line 63: please format changes

src/test/java/org/folio/rest/impl/MockServer.java
```diff
@@ -494,6 +497,20 @@ private void handleGetUsersRecord(RoutingContext ctx) {
     }
   }
 
+  private void handleGetInstanceBulkIds(RoutingContext ctx) {
+    logger.info("Handle get instance bulk ids: " + ctx.request()
```
Igor-Gorchakov, line 501: in one line please

src/main/java/org/folio/clients/InventoryClient.java
illia-borysenko: I think it's not used import, I will remove it, thanks

ramls/schemas/fileDefinition.json
```diff
@@ -36,6 +36,15 @@
       "description": "Size of the file in Kbyte",
       "type": "integer"
     },
+    "uploadFormat": {
+      "description": "Format of exported file",
```
KVupp, line 40: Please change the description to "type of the upload file"

descriptors/ModuleDescriptor-template.json
```diff
@@ -113,7 +117,8 @@
             "configuration.entries.collection.get",
             "inventory-storage.holdings-note-types.collection.get",
             "inventory-storage.item-note-types.collection.get",
-            "inventory-storage.call-number-types.collection.get"
+            "inventory-storage.call-number-types.collection.get",
+            "inventory-storage.instance-bulk.ids.get"
```
KVupp, line 121: /this permission need to be given to upload endpoint isn't it? That's where you are querying to fetch the UUIDs

Igor-Gorchakov (approved):

descriptors/ModuleDescriptor-template.json
illia-borysenko: yes, it's permission required to retrieve UUIDs by /instance-bulk/ids

KVupp: you added it in the incorrect place Illia, Please add it to the upload endpoint

src/main/java/org/folio/service/file/upload/FileUploadServiceImpl.java
```diff
@@ -43,14 +56,36 @@ public FileUploadServiceImpl(@Autowired FileStorage fileStorage, @Autowired File
...
+        JobExecution jobExecution = new JobExecution().
+          withProgress(new Progress()
+            .withTotal(String.valueOf(jsonIds.size())));
+        return fileStorage.saveFileDataAsyncCQL(ids, fileDefinition)
```
KVupp, line 84: How is this process asynchronous? You are waiting until the file is stored on disk using a blocking operation

illia-borysenko: yes it's true, nevetheless there is promise inside ant it returns future, this method the same as saveFileDataAsync, but we also have saveFileDataBlocking that returns FileDefinition

src/test/resources/files/InventoryUUIDs.cql
```diff
@@ -0,0 +1 @@
+(source=="FOLIO") sortby title
```
KVupp, line 1: This is a duplicate file, same as above

src/test/java/org/folio/rest/impl/DataExportTest.java
```diff
@@ -152,23 +165,54 @@ void testExport_GenerateRecordsOnFly(VertxTestContext context) throws IOExceptio
...
+  }
+
+  @Test
+  void testExportByCQL_GenerateRecordsOnFly_andUnderlyingSrs(VertxTestContext context) throws IOException {
+    // given
+    String tenantId = okapiConnectionParams.getTenantId();
+    JsonObject bulkUUIDs = new JsonObject().put("ids", new JsonArray()
+      .add("7fbd5d84-62d1-44c6-9c45-6cb173998bbd")
+      .add("5fc04e92-70dd-46b8-97ea-194015762a60"));
+    Mockito.doReturn(Optional.of(bulkUUIDs)).when(mockInventoryClient).getInstancesBulkUUIDs(anyString(), any(OkapiConnectionParams.class));
```
KVupp, line 185: Why do you need to mock this? You already have the call specified in mock server, that should be able to return the values

illia-borysenko: We don't use MockServer in this test

KVupp: We do, it extends the "RestVerticleTestBase" which automatically inititates the mock server. 
Also if you look at validateExternalCalls() , we validate the calls that are inititaed from mock server
This class is for testing end-end, so the actual idea is not to mock the calls here but rather have it tested end-end

src/test/resources/files/InventoryUUIDs.cql
illia-borysenko: it's because for DataExportTest and FileUploadServiceTest we use different directories to get files, it's easier to have two the same files than change this mechanism

PR #178, opened 2020-10-12
MDEXP-267: add progress object to jobExecution when uploading an empty file

IlliaDaliek:
###**PURPOSE**
Add progress object to jobExecution when uploading an empty file.
Jira: https://issues.folio.org/browse/MDEXP-267

### DIFF
--- a/src/test/java/org/folio/rest/impl/DataExportTest.java
+++ b/src/test/java/org/folio/rest/impl/DataExportTest.java
@@ -117,26 +117,6 @@ class DataExportTest extends RestVerticleTestBase {
         });
       }));
   }
-//
-//  @Test
-//  void testExport_uploadingEmptyCsvFile_FAILED_job(VertxTestContext context) throws IOException {
-//    //given
-//    String tenantId = okapiConnectionParams.getTenantId();
-//    FileDefinition uploadedFileDefinition = uploadFile(EMPTY_FILE, CSV, buildRequestSpecification(tenantId));
-//    // when
-//    ExportRequest exportRequest = buildExportRequest(uploadedFileDefinition);
-//    postRequest(JsonObject.mapFrom(exportRequest), EXPORT_URL);
-//    String jobExecutionId = uploadedFileDefinition.getJobExecutionId();
-//    // then
-//    vertx.setTimer(TIMER_DELAY, handler ->
-//      jobExecutionDao.getById(jobExecutionId, tenantId).onSuccess(optionalJobExecution -> {
-//        JobExecution jobExecution = optionalJobExecution.get();
-//        context.verify(() -> {
-//          assertJobExecution(jobExecution, FAIL, EXPORTED_RECORDS_EMPTY);
-//          context.completeNow();
-//        });
-//      }));
-//  }
 
   @Test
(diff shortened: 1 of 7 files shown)


ViktorSoroka07:
Please name commits with placing the number of ticket e.g `MDEXP-267: ` in the beginning so they propagate to JIRA.

KVupp:
> Please name commits with placing the number of ticket e.g `MDEXP-267: ` in the beginning so they propagate to JIRA.

@ViktorSoroka07, In backend we squash and merge commits, so we **don't need** to prefix the ticket number of each commit. We just need the PR needs to be prefixed with ticket number, and the final merge commit to contain it.

src/test/java/org/folio/rest/impl/DataExportTest.java
```diff
@@ -86,6 +90,26 @@ public DataExportTest() {
...
+    // when
+    ExportRequest exportRequest = buildExportRequest(uploadedFileDefinition);
+    postRequest(JsonObject.mapFrom(exportRequest), EXPORT_URL);
+    String jobExecutionId = uploadedFileDefinition.getJobExecutionId();
+    // then
+    vertx.setTimer(TIMER_DELAY, handler ->
+      jobExecutionDao.getById(jobExecutionId, tenantId).onSuccess(optionalJobExecution -> {
+        JobExecution jobExecution = optionalJobExecution.get();
+        context.verify(() -> {
+          assertJobExecution(jobExecution, FAIL, EXPORTED_RECORDS_EMPTY);
```
KVupp, line 107: Why is it asserted to null value? It must be 0, if you are creating a new Progress Object

ViktorSoroka07:
> @ViktorSoroka07, In backend we squash and merge commits, so we don't need to prefix the ticket number of each commit. We just need the PR needs to be prefixed with ticket number, and the final merge commits to contain it.

At least one commit should be with the ticket number while creating the PR, so there is a link in JIRA to the PR itself. I find it inconvenient when I see when the ticket is in code review status and then heading over the JIRA don't see the link to the PR. In UI we name every commit with the prefix (it does not really hurt) and we squash them too.

src/test/java/org/folio/rest/impl/DataExportTest.java
andrii-novytskyi: In my opinion, Progress should be created as a new object, and all fields should be 0

andrii-novytskyi:
@IlliaDaliek Please, upgrade the test coverage at least to 80% is possible

src/test/java/org/folio/rest/impl/DataExportTest.java
IlliaDaliek: Constant field value has been changed to 0 value

illia-borysenko (approved):

src/test/java/org/folio/rest/impl/DataExportTest.java
andrii-novytskyi: @IlliaDaliek could you please test if csv file is empty

andrii-novytskyi (approved):

PR #185, opened 2020-10-19
MDEXP-267: fix jobExecution empty fileName and runBy when uploading an empty file.

IlliaDaliek:
## Purpose
Add fileName and runBy fields to jobExecution when uploading an empty file for borh csv and cql upload formats.
Jira: https://issues.folio.org/browse/MDEXP-267
results

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -474,7 +474,7 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-surefire-plugin</artifactId>
-        <version>2.22.1</version>
+        <version>2.22.2</version>
         <configuration>
           <useSystemClassLoader>false</useSystemClassLoader>
           <systemPropertyVariables>
--- a/src/main/java/org/folio/service/manager/input/InputDataManagerImpl.java
+++ b/src/main/java/org/folio/service/manager/input/InputDataManagerImpl.java
@@ -1,14 +1,16 @@
 package org.folio.service.manager.input;
 
-import io.vertx.core.AsyncResult;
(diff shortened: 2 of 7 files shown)


pom.xml
```diff
@@ -474,7 +474,7 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-surefire-plugin</artifactId>
-        <version>2.22.1</version>
+        <version>2.22.2</version>
```
IlliaDaliek, line 477: I increased the version cause it's more stable unlike 2.22.1 and I can't get coverage info and successful pre-merge build with 2.22.1

src/main/java/org/folio/service/manager/input/InputDataManagerImpl.java
```diff
@@ -255,6 +260,24 @@ private boolean isNotCQL(FileDefinition requestFileDefinition) {
     return !CQL.equals(requestFileDefinition.getUploadFormat());
   }
 
+  private void populateJobExecutionForEmptyFileUploaded(JobExecution jobExecution, FileDefinition fileExportDefinition, JsonObject user) {
```
andrii-novytskyi, line 263: the method looks very similar to the prepareJobForExport(). Maybe. we can move a common part to one place

IlliaDaliek: Thank for note, I've added changes regarding your suggestion.

Igor-Gorchakov (approved):

illia-borysenko (approved):

PR #192, opened 2020-10-23
MDEXP-305 Fix generated mrc files and directories are not being deleted

andrii-novytskyi:
## Purpose
https://issues.folio.org/browse/MDEXP-305

## Approach
While investigating the issue, I saw that it is unreproducible on test/scratch envs. Investigated logs from bugfest-goldenroad, no logs about cleanup process - so the cleanup job is not running at all, or stuck. 
Decided to change mechanism of the cleanup process to provide a new endpoint, and use okapi timer to retrieve this API.
Also, it will give us the opportunity to manually start the cleanup process.
Updated rancher configs to helm configs with version: 0.1.30

### DIFF
--- a/src/main/java/org/folio/rest/impl/ModTenantAPI.java
+++ b/src/main/java/org/folio/rest/impl/ModTenantAPI.java
@@ -29,9 +29,7 @@ public class ModTenantAPI extends TenantAPI {
   public void postTenant(TenantAttributes entity, Map<String, String> headers, Handler<AsyncResult<Response>> handlers, Context context) {
     super.postTenant(entity, headers, asyncResult -> {
       if (asyncResult.failed()) {
-        if (Objects.nonNull(asyncResult.cause())) {
-          LOGGER.error("Post tenant is failed, cause: " + asyncResult.cause().getMessage());
-        }
+        LOGGER.error("Post tenant is failed, cause: " + asyncResult.cause().getMessage());
         handlers.handle(asyncResult);
       } else {
         LOGGER.info("Post tenant is complete successfully");
--- a/src/main/java/org/folio/rest/impl/ModTenantAPI.java
+++ b/src/main/java/org/folio/rest/impl/ModTenantAPI.java
@@ -12,7 +12,6 @@ import org.folio.spring.SpringContextUtil;
 
 import javax.ws.rs.core.Response;
 import java.util.Map;
-import java.util.Objects;
 
 import static io.vertx.core.Future.succeededFuture;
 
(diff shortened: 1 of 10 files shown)


andrii-novytskyi:
@folio-org/dataexportbackend The rest assured tests are present in PR, but for some reason sonar thinking that new method in controller not covered, does anybody knows what is the reason?

src/main/java/org/folio/rest/impl/ModTenantAPI.java
```diff
@@ -35,28 +29,13 @@ public ModTenantAPI() { //NOSONAR
   public void postTenant(TenantAttributes entity, Map<String, String> headers, Handler<AsyncResult<Response>> handlers, Context context) {
     super.postTenant(entity, headers, asyncResult -> {
       if (asyncResult.failed()) {
+        LOGGER.error("Post tenant is failed, cause: " + asyncResult.cause().getMessage());
```
KVupp, line 32:
```suggestion
        LOGGER.error("Post tenant failed, cause: " + asyncResult.cause().getMessage());
```

src/main/java/org/folio/rest/impl/ModTenantAPI.java
```diff
@@ -35,28 +29,13 @@ public ModTenantAPI() { //NOSONAR
...
         handlers.handle(asyncResult);
       } else {
-        initStorageCleanupService(headers, context);
+        LOGGER.info("Post tenant is complete successfully");
```
KVupp, line 35:
```suggestion
        LOGGER.info("Post tenant is completed successfully");
```

src/main/java/org/folio/service/file/cleanup/StorageCleanupServiceImpl.java
```diff
@@ -44,6 +46,7 @@ private Date getFileDefinitionExpirationDate() {
   }
 
   private Future<CompositeFuture> deleteExpiredFilesAndRelatedFileDefinitions(List<FileDefinition> fileDefinitions, String tenantId) {
+    LOGGER.info(format("Start process of removing files and file definitions, number of file definitions to clean up: %s", fileDefinitions.size()));
```
KVupp, line 49: Ideally we should be using the parameterized logging offered by log4j : https://logging.apache.org/log4j/2.x/performance.html#asyncLoggingWithParams

andrii-novytskyi: @KVupp thanks, done

KVupp (approved):

Igor-Gorchakov (approved):

PR #193, opened 2020-10-29
MDEXP-308 Subfield $3 is missing for the MARC tags with item data

Igor-Gorchakov:
https://issues.folio.org/browse/MDEXP-308
Linked PR https://github.com/folio-org/generate-marc-utils/pull/20
## Purpose
The purpose is to correct the previous behavior of how the RuleFactory.java was working.
Before: the RuleFactory marks Item transformations that have a common marc tag with the parent Holding transformation
Now: the RuleFactory marks only Item transformations that have a common marc tag with other Item transformations
## Approach
-- Fixed RuleFactory
-- Factored RuleHandler, using Rule.isHasSameTagInItems field to identify rules that should be indexed from the original version
_**See verification below**_


Igor-Gorchakov:
I have already verified these changes with Magda, demoed on the localhost.
Verification:
**1. Create a mapping profile**
image

**2. Given Inventory Instance has 2 holdings, 1st Holding has 2 Items,  2nd Holding has 1 Item**
image

**3. Generated marc file**
image

illia-borysenko (approved):

ramls/schemas/profiles/transformations.json
```diff
@@ -27,8 +27,8 @@
       "type": "string",
       "$ref": "recordTypes.json"
     },
-    "hasSameTagInHoldings": {
-      "description": "The sign that means the current transformation intended for both item and holding",
+    "hasSameTagInItems": {
```
KVupp, line 30: This is still a breaking change, I don't know how this PR is different from the previous one, but we still need to discuss it. This will need a migration script

andrii-novytskyi: @KVupp ideally, this can be not broken changes, as this field is only for backend usage, it calculates at the backend, and used there, we don`t need it on UI side, or in the DB. Just need to ensure that these fields are not present in the DB (on the rancher, maybe someone created a mapping profile by the postman, and add this field in transformation, that`s why we had those issue). But we need to ensure that this field does not get to the DB, if present there - then this is a breaking change in my opinion

Igor-Gorchakov: I will reach Rancher to see mapping profiles, I don't think this field can get the DB

Igor-Gorchakov: @KVupp  @andrii-novytskyi the script is in place

andrii-novytskyi (approved):

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -109,6 +109,11 @@
       "run": "after",
       "snippetPath": "migration_scripts/change_progress_total_data_type.sql",
       "fromModuleVersion": "mod-data-export-3.1.0"
+    },
+    {
+      "run": "after",
+      "snippetPath": "migration_scripts/remove_hasSameTagInHoldings.sql",
+      "fromModuleVersion": "mod-data-export-3.1.0"
```
KVupp, line 116: we'll release this for bugfix release, so version will be 3.0.1

Igor-Gorchakov: updated

ramls/schemas/profiles/transformations.json
```diff
@@ -27,8 +27,8 @@
...
+      "description": "The sign that means the current transformation has same marc tag in other item transformations",
```
KVupp, line 31:
```suggestion
      "description": "The field indicates that the current transformation has same marc tag in other item transformations",
```

KVupp (approved):

PR #195, opened 2020-10-30
MDEXP-311 Fix wrong job status when all UUIDs in the last batch not present in storages

andrii-novytskyi:
## Purpose
https://issues.folio.org/browse/MDEXP-311

## Approach
If the batch is last and no records are found in inventory or srs, add check if mrc file exist in file system, if exist - status CompletedWithErrors, if not - Failed.
Added unit test


Igor-Gorchakov (approved):

src/main/java/org/folio/service/manager/export/ExportManagerImpl.java
```diff
@@ -203,19 +203,26 @@ private ExportResult getExportResult(AsyncResult<Object> asyncResult, ExportPayl
...
-        } else {
-          return ExportResult.completed();
-        }
+        return getProperExportResultIfBatchIsLast(exportPayload);
       }
       return ExportResult.inProgress();
     }
   }
 
+  private ExportResult getProperExportResultIfBatchIsLast(ExportPayload exportPayload) {
```
Igor-Gorchakov, line 212: Rename to getExportResultForLastBatch

src/main/java/org/folio/service/manager/export/ExportManagerImpl.java
```diff
@@ -203,19 +203,26 @@ private ExportResult getExportResult(AsyncResult<Object> asyncResult, ExportPayl
...
+    if (exportPayload.getExportedRecordsNumber() == 0) {
+      if (vertx.fileSystem().existsBlocking(exportPayload.getFileExportDefinition().getSourcePath())) {
```
Igor-Gorchakov, line 214: Use FileStorage service to check the existence of file

src/main/java/org/folio/service/manager/export/ExportManagerImpl.java
andrii-novytskyi: done

src/main/java/org/folio/service/manager/export/ExportManagerImpl.java
andrii-novytskyi: done

IlliaDaliek (approved):

PR #198, opened 2020-11-05
MDEXP-316 Fix error logs doesn`t contain item info

andrii-novytskyi:
## Purpose
https://issues.folio.org/browse/MDEXP-316

## Approach
Since the structure of the JsonObject that goes into the mapping process has changed, we have to change the creation of AffectedRecords, so we added changes to this PR.

### DIFF
--- a/src/test/java/org/folio/service/logs/ErrorLogServiceUnitTest.java
+++ b/src/test/java/org/folio/service/logs/ErrorLogServiceUnitTest.java
@@ -8,7 +8,6 @@ import io.vertx.ext.unit.junit.VertxUnitRunner;
 import io.vertx.junit5.VertxExtension;
 import io.vertx.junit5.VertxTestContext;
 import org.apache.commons.collections4.map.HashedMap;
-import org.apache.commons.lang3.StringUtils;
 import org.assertj.core.util.Lists;
 import org.folio.dao.impl.ErrorLogDaoImpl;
 import org.folio.rest.jaxrs.model.AffectedRecord;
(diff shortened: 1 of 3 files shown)


src/test/java/org/folio/service/logs/ErrorLogServiceUnitTest.java
```diff
@@ -207,52 +205,14 @@ void saveWithAffectedRecord_shouldSaveWithHoldingsAndItemsInfo(VertxTestContext
...
-      assertTrue(ar.succeeded());
-      verify(errorLogDao).save(errorLogCaptor.capture(), eq(TENANT_ID));
-      ErrorLog errorLog = errorLogCaptor.getValue();
-      assertErrorLogWithHoldingsAndItems(errorLog, false);
-      context.completeNow();
-    }));
-  }
-
-  @Test
-  void saveWithAffectedRecord_shouldSaveWithHoldingsInfoWithoutItemIfHoldingIdIsEmpty(VertxTestContext context) {
```
KVupp, line 226: WHy is this test case removed, it should still pass right?

KVupp (approved):
Just a minor comment regarding the removed test cases. Otherwise LGTM

src/test/java/org/folio/service/logs/ErrorLogServiceUnitTest.java
andrii-novytskyi: removed check when item holdingRecordId equals to holding, as we do it in fetchHoldingsAndItems() method while building json for mapping process, so we don`t need this test

KVupp:
As part of this ticket can you also please increase the RMB version that was recommended to 31.1.5

IlliaDaliek (approved):

PR #202, opened 2020-11-11
MDEXP-337 Fix correct job status when random uuids not in last batch

andrii-novytskyi:
## Purpose
When the random UUIDs not in the last batch - then job execution status always completed, instead of completed with errors.
Also, if the CSV file contains some text, not UUIDs, status should be completed with errors too.

## Approach
After calling finalize export, check if the error logs related to UUIDs are present, if yes - status completed with errors, if not - completed.

### DIFF
--- a/src/main/java/org/folio/service/manager/input/InputDataManagerImpl.java
+++ b/src/main/java/org/folio/service/manager/input/InputDataManagerImpl.java
@@ -171,12 +171,12 @@ class InputDataManagerImpl implements InputDataManager {
     if (status.equals(JobExecution.Status.COMPLETED)) {
       isErrorsRelatedToUUIDsPresent(jobExecutionId, tenantId)
           .onComplete(
-            isAnyErrorPresent -> {
-                if (isAnyErrorPresent.succeeded() && Boolean.TRUE.equals(isAnyErrorPresent.result())) {
+              isAnyErrorPresent -> {
+                if (isAnyErrorPresent.succeeded()
+                    && Boolean.TRUE.equals(isAnyErrorPresent.result())) {
                   exportResult.setStatus(COMPLETED_WITH_ERRORS);
                 }
-                jobExecutionService.updateJobStatusById(
-                    jobExecutionId, getJobExecutionStatus(exportResult), tenantId);
+                jobExecutionService.updateJobStatusById(jobExecutionId, getJobExecutionStatus(exportResult), tenantId);
                 updateFileDefinitionStatusByResult(fileExportDefinition, exportResult, tenantId);
               });
     } else {
--- a/src/main/java/org/folio/service/manager/input/InputDataManagerImpl.java
+++ b/src/main/java/org/folio/service/manager/input/InputDataManagerImpl.java
@@ -1,14 +1,16 @@
 package org.folio.service.manager.input;
 
-import static io.vertx.core.Future.succeededFuture;
-import static java.util.Objects.nonNull;
-import static org.folio.rest.jaxrs.model.FileDefinition.UploadFormat.CQL;
-
-import java.lang.invoke.MethodHandles;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-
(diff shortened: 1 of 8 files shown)


src/main/java/org/folio/service/manager/input/InputDataManagerImpl.java
```diff
@@ -165,12 +168,49 @@ private void finalizeExport(ExportPayload exportPayload, ExportResult exportResu
...
     removeInputDataContext(jobExecutionId);
   }
 
+  private Future<Boolean> isErrorsRelatedToUUIDsPresent(String jobExecutionId, String tenantId) {
+    Promise<Boolean> promise = Promise.promise();
+    List<Future> errorsList =
+        ErrorCode.reasonsAccordingToUUIDs().stream()
+            .map(
+                errorCode ->
+                    errorLogService.isErrorsByReasonPresent(errorCode, jobExecutionId, tenantId))
```
KVupp, line 197: so for each reason we are querying the DB, I think that will be a lot of DB calls, we can do a "IN" query instead, that way only 1 DB call is made

src/main/java/org/folio/service/manager/input/InputDataManagerImpl.java
```diff
@@ -165,12 +168,49 @@ private void finalizeExport(ExportPayload exportPayload, ExportResult exportResu
     FileDefinition fileExportDefinition = exportPayload.getFileExportDefinition();
     String jobExecutionId = fileExportDefinition.getJobExecutionId();
     String tenantId = exportPayload.getOkapiConnectionParams().getTenantId();
-    jobExecutionService.updateJobStatusById(jobExecutionId, getJobExecutionStatus(exportResult), tenantId);
-    updateFileDefinitionStatusByResult(fileExportDefinition, exportResult, tenantId);
+    JobExecution.Status status = getJobExecutionStatus(exportResult);
```
KVupp, line 171: Also the getJobExecutionStatus is likely not necessary right? All that is doing is checking the status based on the last export that happened. Which is not required because we are deciding it based on the errors.

andrii-novytskyi: we need this status to check if it is completed. then new logic will be called. If status not completed, we don`t need it, cause error logs are present and user will able to open it, and we don`t need to do a call to the DB

KVupp: ok yea makes sense, in case the last batch has failed or completed with error we save a DB call

illia-borysenko (approved):

src/main/java/org/folio/service/manager/input/InputDataManagerImpl.java
andrii-novytskyi: updated with using `similar to` param

src/main/java/org/folio/service/manager/input/InputDataManagerImpl.java
```diff
@@ -165,12 +167,36 @@ private void finalizeExport(ExportPayload exportPayload, ExportResult exportResu
...
+    if (status.equals(JobExecution.Status.COMPLETED)) {
+      isErrorsRelatedToUUIDsPresent(jobExecutionId, tenantId)
+          .onComplete(
+            isAnyErrorPresent -> {
+                if (isAnyErrorPresent.succeeded() && Boolean.TRUE.equals(isAnyErrorPresent.result())) {
+                  exportResult.setStatus(COMPLETED_WITH_ERRORS);
+                }
+                jobExecutionService.updateJobStatusById(
+                    jobExecutionId, getJobExecutionStatus(exportResult), tenantId);
```
Igor-Gorchakov, line 179: Google code style looks strange, please keep it in one line

Igor-Gorchakov (approved):

KVupp:
Please take a note of the sonar errors before merging. Thanks!

PR #221, opened 2020-12-09
MDEXP-351 Add check to not store MARC record in file if data is empty

andrii-novytskyi:
## Purpose
https://issues.folio.org/browse/MDEXP-351

## Approach
Add check to not store file if data is empty during export inventory||srs records. We will avoid creating a file and storing it in S3
with empty data. 

Related PR: https://github.com/folio-org/generate-marc-utils/pull/29

### DIFF
--- a/src/test/java/org/folio/service/export/ExportServiceUnitTest.java
+++ b/src/test/java/org/folio/service/export/ExportServiceUnitTest.java
@@ -1,24 +1,10 @@
 package org.folio.service.export;
 
-import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.ArgumentMatchers.anyString;
-import static org.mockito.ArgumentMatchers.eq;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
-
-import com.fasterxml.jackson.annotation.ObjectIdGenerators;
 import io.vertx.core.json.JsonObject;
-
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.List;
-import java.util.UUID;
-
+import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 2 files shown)


src/test/java/org/folio/service/export/ExportServiceUnitTest.java
```diff
@@ -92,6 +84,29 @@ void shouldPassExportFor_1_InventoryRecord() {
     Mockito.verify(fileStorage, Mockito.times(1)).saveFileDataBlocking(any(byte[].class), any(FileDefinition.class));
   }
 
+  @Test
+  void shouldNotStoreFile_whenInventoryRecordIsEmpty() {
```
KVupp, line 88:
```suggestion
  void shouldNotStoreRecord_whenInventoryRecordIsEmpty() {
```

src/test/java/org/folio/service/export/ExportServiceUnitTest.java
```diff
@@ -92,6 +84,29 @@ void shouldPassExportFor_1_InventoryRecord() {
...
+    String inventoryRecord = StringUtils.EMPTY;
+    FileDefinition fileDefinition = new FileDefinition();
+    // when
+    exportService.exportInventoryRecords(Collections.singletonList(inventoryRecord), fileDefinition, TENANT);
+    // then
+    Mockito.verify(fileStorage, never()).saveFileDataBlocking(any(byte[].class), any(FileDefinition.class));
+  }
+
+  @Test
+  void shouldNotStoreFile_whenSRSRecordIsEmpty() {
```
KVupp, line 99:
```suggestion
  void shouldNotStoreMarcInFile_whenSRSRecordIsEmpty() {
```

KVupp (approved):

PR #222, opened 2020-12-10
MDEXP-344/348: Allow to pass empty transformations for all record types except ITEM type, allow to have characters in indicator fields.

IlliaDaliek:
Jira: 
- https://issues.folio.org/browse/MDEXP-344
- https://issues.folio.org/browse/MDEXP-348

## Purpose
Allow to pass an empty transformation for INSTANCE and HOLDING record types only, as well allow to have characters in indicator fields.

## Approach
Move validation from RMB side to service layer logic.

## Pre-Merge Checklist


src/test/java/org/folio/rest/impl/MappingProfileServiceTest.java
```diff
@@ -14,12 +12,23 @@
...
 @ExtendWith(VertxExtension.class)
 class MappingProfileServiceTest extends RestVerticleTestBase {
 
   private static final String INVALID_TRANSFORMATION = "123qq$qweqwe";
+  private static final String INVALID_TAG = "12a  $a";
+  private static final String INVALID_FIRST_INDICATOR = "123. $a";
+  private static final String INVALID_SECOND_INDICATOR = "123 ,$a";
+  private static final String INVALID_SUBFIELD = "123  $1234";
+  private static final String EMPTY_TRANSFORMATION = "";
+  private static final String VALID_TRANSFORMATION = "13qq$q";
```
andrii-novytskyi, line 31: unused field

src/main/java/org/folio/service/profiles/mappingprofile/MappingProfileServiceImpl.java
```diff
@@ -123,31 +125,35 @@
...
-          TransformationField.RecordType expectedRecordType = transformationField.getRecordType();
-          if (Objects.isNull(transformation.getRecordType()) || !transformation.getRecordType().toString().equals(expectedRecordType.toString())) {
-            throw new ServiceException(HttpStatus.HTTP_UNPROCESSABLE_ENTITY, String.format("Transformation record type is missing or incorrect according to provided fieldId: %s, " +
-              "expected record type: %s", fieldId, expectedRecordType));
-          }
-        }
-        return Future.succeededFuture();
-      });
+      transformationFieldsService.validateTransformations(mappingProfile.getTransformations())
+        .compose(v -> transformationFieldsService.getTransformationFields(params)
```
andrii-novytskyi, line 131: why you use compose, and then onSuccess on tris compose, and on failure. Probably, it is better to use  onSuccess and on failure only if you use it.
Also, you forgot to complete promise if mappingProfile.getTransformations() is empty, that s why your tests are failed

illia-borysenko (approved):

src/main/java/org/folio/service/profiles/mappingprofile/MappingProfileServiceImpl.java
```diff
@@ -123,31 +123,52 @@
...
   @Override
   public Future<Void> validate(MappingProfile mappingProfile, OkapiConnectionParams params) {
+    Promise<Void> promise = Promise.promise();
     if (CollectionUtils.isNotEmpty(mappingProfile.getTransformations())) {
-      return transformationFieldsService.getTransformationFields(params).compose(transformationFieldCollection -> {
-        List<TransformationField> transformationFields = transformationFieldCollection.getTransformationFields();
-        for (Transformations transformation : mappingProfile.getTransformations()) {
+      transformationFieldsService.validateTransformations(mappingProfile.getTransformations())
+        .compose(v -> transformationFieldsService.getTransformationFields(params)
+          .compose(transformationFieldCollection -> validateTransformation(mappingProfile, transformationFieldCollection)))
```
KVupp, line 130: We should combine all validations to a single place, it's difficult to keep track of so many validate methods

ramls/schemas/profiles/transformations.json
```diff
@@ -20,8 +20,7 @@
     },
     "transformation": {
       "description": "Mapping expression",
-      "type": "string",
-      "pattern": "(\\d{3}[\\s\\d]{2}(\\$([a-zA-Z]|[\\d]{1,2}))?)"
+      "type": "string"
```
KVupp, line 23: Just a thought, we should have left the pattern and just added an empty to the pattern basically like UUID || "".
That way in this method only if part is needed, the else part can be removed totally :https://github.com/folio-org/mod-data-export/pull/222/files#diff-dc8e8e8fd68ceeeb43c5e09450d6a1ac227847835f12cfcbc14f07645d151479R78

src/main/java/org/folio/service/profiles/mappingprofile/MappingProfileServiceImpl.java
```diff
@@ -123,31 +123,52 @@
...
+    }
+    return succeededFuture();
+  }
+
+  private Optional<TransformationField> getTransformationField(List<TransformationField> transformationFields, Transformations transformation) {
+    Optional<TransformationField> transformationFieldOptional = transformationFields.stream()
+      .filter(transformationField -> transformation.getFieldId().equals(transformationField.getFieldId()))
+      .findFirst();
+    if (transformationFieldOptional.isEmpty()) {
+      throw new ServiceException(HttpStatus.HTTP_UNPROCESSABLE_ENTITY, String.format("Transformation doesn't exist by provided fieldId: %s", transformation.getFieldId()));
```
KVupp, line 169:
```suggestion
      throw new ServiceException(HttpStatus.HTTP_UNPROCESSABLE_ENTITY, String.format("Transformation doesn't exist for provided fieldId: %s", transformation.getFieldId()));
```

src/main/java/org/folio/service/transformationfields/TransformationFieldsServiceImpl.java
```diff
@@ -52,6 +70,50 @@
     return promise.future();
   }
 
+  @Override
+  public Future<Void> validateTransformations(List<Transformations> transformations) {
```
KVupp, line 74: Also please add some docs here, as to what validations this method does, for readability

IlliaDaliek: I added Javadoc, thank you

ramls/schemas/profiles/transformations.json
IlliaDaliek: Yes, only by the reason that we cannot allow saving empty values for Item record type we should do all this validation at the service layer.

src/main/java/org/folio/service/profiles/mappingprofile/MappingProfileServiceImpl.java
IlliaDaliek: I think that this one was simplified well. Logic is the same. Wath do you think guys @illia-borysenko , @Igor-Gorchakov , @andrii-novytskyi ?

src/test/java/org/folio/rest/impl/MappingProfileServiceTest.java
IlliaDaliek: Corrected, I've replaced the old variable with a new one which is used.

src/main/java/org/folio/service/profiles/mappingprofile/MappingProfileServiceImpl.java
andrii-novytskyi: we can create a separate class for validation purpose. @KVupp did you mean that?

src/main/java/org/folio/service/profiles/mappingprofile/MappingProfileServiceImpl.java
IlliaDaliek: Fixed

src/main/java/org/folio/service/profiles/mappingprofile/MappingProfileServiceImpl.java
KVupp: yea there are multiple validation methods and all over the place, we need to do the validation only in one place, move the other code along with it, so it's more readable

ramls/schemas/profiles/transformations.json
KVupp: right only the item validation can be handled in the service layer, pattern can still be done here right

IlliaDaliek: So, how are you going to apply RMB validation only for Instance and Holding record types? This pattern setting into json file is global, I mean it is applied always, no matter which any other fields passed. Correct me please if I am wrong.

KVupp: RMB will validate the pattern (a regex which will allow either UUID or an empty string) .
In the validate step all we have to do is to check if **item** transformations are empty or not. WE need to make this change now. But just mentioning that it would have made the code simpler

IlliaDaliek: > RMB will validate the pattern (a regex which will allow either UUID or an empty string) .
> In the validate step all we have to do is to check if **item** transformations are empty or not. WE need to make this change now. But just mentioning that it would have made the code simpler
We cannot do validation at RMB level because we should not allow empty values for transformation in the case of the Item record type. Therefore we should check the record type at the service level, we cannot do this at RMB level.

src/main/java/org/folio/service/profiles/mappingprofile/MappingProfileServiceImpl.java
KVupp: We decided to handle in a future PR

ramls/schemas/profiles/transformations.json
KVupp: Much cleaner now. Thanks for the changes

src/test/java/org/folio/service/mapping/MappingServiceUnitTest.java
```diff
@@ -251,7 +269,7 @@ void shouldReturnVariableFieldsForHoldingsAndItem_whenMappingProfileTransformati
     Assert.assertEquals(38, appendedMarcRecords.stream().map(vf -> vf.getTag()).collect(Collectors.toSet()).size());
     Assert.assertEquals(41, appendedMarcRecords.size());
   }
-
+//kek
```
KVupp, line 272: typo?

IlliaDaliek: yes

KVupp (approved):

PR #225, opened 2020-12-14
MDEXP-349 Change quick export endpoint to return jobExecution id and hrId, add fileName to request

andrii-novytskyi:
## Purpose
https://issues.folio.org/browse/MDEXP-349

## Approach
1. Return a 200 response that includes the jobExecutionId and jobExecution.hrid in the body;
2.  Add an optional `fileName` key to the request body. If fileName is not present, the default  **quick-export-HRID** name will be used

### DIFF
--- a/ramls/schemas/quickExportRequest.json
+++ b/ramls/schemas/quickExportRequest.json
@@ -37,6 +37,10 @@
         "ITEM"
       ]
     },
+    "fileName": {
+      "description": "Name of the file to export",
+      "type": "string"
+    },
     "metadata": {
       "description": "Meta information ",
       "type": "object",
--- a/ramls/schemas/quickExportResponse.json
+++ b/ramls/schemas/quickExportResponse.json
@@ -1,0 +1,20 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "description": "Quick export response",
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+    "jobExecutionId": {
+      "description": "UUID of the job execution, that is used for quick export",
+      "type": "string"
+    },
+    "jobExecutionHrId": {
+      "description": "HrId of the job execution, that is used for quick export",
+      "type": "integer"
+    }
+  },
+  "required": [
+    "jobExecutionId",
+    "jobExecutionHrId"
+  ]
+}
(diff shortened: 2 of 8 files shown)


illia-borysenko (approved):

IlliaDaliek (approved):

ramls/schemas/quickExportResponse.json
```diff
@@ -0,0 +1,20 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "description": "Quick export response",
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+    "jobExecutionId": {
+      "description": "UUID of the job execution, that is used for quick export",
+      "type": "string"
```
KVupp, line 9: Not a big thing but why didn't we add the UUID like we usually do with pattern or infact we already have it in jobExecution.json?

src/main/java/org/folio/service/file/definition/FileDefinitionServiceImpl.java
```diff
@@ -84,4 +85,10 @@
       : FileDefinition.UploadFormat.CSV;
   }
 
+  private String getFileNameByRequest(QuickExportRequest request) {
+    return StringUtils.isEmpty(request.getFileName())
+      ? QUICK_EXPORT + CSV_FILE_FORMAT
+      : request.getFileName() + CSV_FILE_FORMAT;
```
KVupp, line 91: what if the file name given is test.csv?  we'll be appending csv twice?

ramls/schemas/quickExportRequest.json
```diff
@@ -37,6 +37,10 @@
         "ITEM"
       ]
     },
+    "fileName": {
+      "description": "Name of the file to export",
```
KVupp, line 41: Isn't it actually the name of the marc file, infact it is the file name to be downloaded. We should name it appropriately or atleast give a clear description

KVupp (commented):
Sorry to have reviewed it after the merge. But just left a few minor comments

src/main/java/org/folio/service/file/definition/FileDefinitionServiceImpl.java
KVupp: or infact because this is the filename of download what if user gives test.mrc?

ramls/schemas/quickExportResponse.json
andrii-novytskyi: we already have it in jobExecution, but if you want I will add it in the scope of my current story

src/main/java/org/folio/service/file/definition/FileDefinitionServiceImpl.java
andrii-novytskyi: we expect fileName, not fileName with extension. If the user will provide such a name, for example, test.mrc, the result will be 
image

ramls/schemas/quickExportRequest.json
andrii-novytskyi: I will rename description in my current story, thanks

PR #250, opened 2021-07-06
MDEXP-402: Migration Script does not handle all cases

IlliaDaliek:
Jira: https://issues.folio.org/browse/MDEXP-402

### PURPOSE
Enhance migration script with additional checks that nested 'progress' object is not empty and populate existed empty 'progress' objects with 0 values for all possible nested fields like total, failed, and exported. Cover the case where only the 'exported' field is presented and populate such 'progress' objects with total field = exported field value and failed = 0.

### DIFF
--- a/src/main/resources/templates/db_scripts/migration_scripts/change_progress_total_data_type.sql
+++ b/src/main/resources/templates/db_scripts/migration_scripts/change_progress_total_data_type.sql
@@ -11,4 +11,4 @@ SET jsonb = jsonb_set(jsonb, '{progress}', jsonb_build_object('total', jsonb -> 
 WHERE jsonb -> 'progress' IS NOT NULL AND jsonb -> 'progress' <> '{}'
 AND jsonb -> 'progress' -> 'exported' IS NOT NULL
 AND jsonb -> 'progress' -> 'total' IS NULL
-AND jsonb -> 'progress' -> 'failed' IS NULL
+AND jsonb -> 'progress' -> 'failed' IS NULL;
--- a/src/main/resources/templates/db_scripts/migration_scripts/change_progress_total_data_type.sql
+++ b/src/main/resources/templates/db_scripts/migration_scripts/change_progress_total_data_type.sql
@@ -2,11 +2,11 @@ UPDATE ${myuniversity}_${mymodule}.job_executions
 SET jsonb = jsonb_set(jsonb #- '{progress, total}', '{progress, total}', to_jsonb((jsonb -> 'progress' ->> 'total')::int))
 WHERE jsonb -> 'progress' IS NOT NULL AND jsonb -> 'progress' <> '{}';
 
-UPDATE diku_mod_data_export.job_executions
+UPDATE ${myuniversity}_${mymodule}.job_executions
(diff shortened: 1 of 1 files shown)


Igor-Gorchakov (approved):

src/main/resources/templates/db_scripts/migration_scripts/change_progress_total_data_type.sql
```diff
@@ -1,3 +1,14 @@
 UPDATE ${myuniversity}_${mymodule}.job_executions
 SET jsonb = jsonb_set(jsonb #- '{progress, total}', '{progress, total}', to_jsonb((jsonb -> 'progress' ->> 'total')::int))
-WHERE jsonb -> 'progress' IS NOT NULL;
+WHERE jsonb -> 'progress' IS NOT NULL AND jsonb -> 'progress' <> '{}';
+
+UPDATE ${myuniversity}_${mymodule}.job_executions
+SET jsonb = jsonb_set(jsonb, '{progress}', '{"total": 0, "failed": 0, "exported": 0}'::jsonb)
+WHERE jsonb -> 'progress' IS NULL OR jsonb -> 'progress' = '{}';
```
mtraneis, line 7: Does the order of these 3 UPDATE statements matter? If the order matter, perhaps this statement should be on top so that it would fill in 0s for the properties, then the last UPDATE  statement below, and lastly the UPDATE statement above to cover all the positive cases, otherwise it would throw errors when certain properties are not present?

mtraneis (commented):
Thanks for working on this. 
Has it been tested with a set of records that did not have all or a combination of the progress object's fields?

src/main/resources/templates/db_scripts/migration_scripts/change_progress_total_data_type.sql
```diff
@@ -1,3 +1,14 @@
...
+
+UPDATE ${myuniversity}_${mymodule}.job_executions
+SET jsonb = jsonb_set(jsonb, '{progress}', jsonb_build_object('total', jsonb -> 'progress' -> 'exported', 'failed', 0, 'exported', jsonb -> 'progress' -> 'exported'))
+WHERE jsonb -> 'progress' IS NOT NULL AND jsonb -> 'progress' <> '{}'
+AND jsonb -> 'progress' -> 'exported' IS NOT NULL
+AND jsonb -> 'progress' -> 'total' IS NULL
+AND jsonb -> 'progress' -> 'failed' IS NULL;
```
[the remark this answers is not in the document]
IlliaDaliek: Hello, @mtraneis . No, it shouldn't. In the system, it is impossible to have such an object that you described. There always will be both exported and total fields.

src/main/resources/templates/db_scripts/migration_scripts/change_progress_total_data_type.sql
IlliaDaliek: @mtraneis , the order doesn't matter here. I tested the scripts thoroughly and there weren'nt any errors. 
I created several jobExecutiuon entities with postgres null, posrtgres ={}, postgres ={exported: some_int}, and progres = {total:"123"}. All entities were processed correctly and I got all entities having the postgres with all 3 fields and total as an integer.

PR #253, opened 2021-07-30
MDEXP-422: Gracefully handle MARC records that exceed the size limit

IlliaDaliek:
Jira: https://issues.folio.org/browse/MDEXP-422

### PURPOSE
Add error handling for JSON records that cannot be converted to marc record for some reason. 

###  APPROACH
Catch en exception, retrieve id from problematic JSON record and obtain related instance by id from inventory-storage. Here the single instance always will be returned(that is why I don't verify the presence of the value of the optional object). Add unit test.


src/main/java/org/folio/service/export/LocalFileSystemExportService.java
```diff
@@ -1,62 +1,114 @@
...
       for (String jsonRecord : jsonRecords) {
-        byte[] bytes = convertJsonRecordToMarcRecord(jsonRecord);
         try {
+          byte[] bytes = convertJsonRecordToMarcRecord(jsonRecord);
           if (isNotEmpty(bytes)) {
             fileStorage.saveFileDataBlocking(bytes, fileDefinition);
           }
+        } catch (MarcException e) {
+          String instId = getInstanceIdFromMarcRecord(new JsonObject(jsonRecord));
+          inventoryClient.getInstancesByIds(Collections.singletonList(instId), jobExecutionId, params, SINGLE_INSTANCE).ifPresent(instancesJson -> {
```
IlliaDaliek, line 79: Here the blocking client is used since all the flow is performed on a separate thread pool and doesn't block the main event loop thread.

src/main/java/org/folio/service/export/LocalFileSystemExportService.java
```diff
@@ -1,62 +1,114 @@
...
+      .filter(jsonObject -> {
+        if (jsonObject.containsKey(GENERAL_INFO_FIELD_TAG_NUMBER)) {
+          JsonObject dataFieldContent = jsonObject.getJsonObject(GENERAL_INFO_FIELD_TAG_NUMBER);
+          String firstIndicator = dataFieldContent.getString(FIRST_INDICATOR);
+          String secondIndicator = dataFieldContent.getString(SECOND_INDICATOR);
+          return StringUtils.isNotBlank(firstIndicator) && StringUtils.isNotBlank(secondIndicator)
+            && firstIndicator.equals(secondIndicator) && firstIndicator.equals(INDICATOR_VALUE);
+        }
+        return false;
+      }).findFirst().get();
```
alekGbuz, line 104: Checking if something exist related conditions may useful here.

IlliaDaliek: @alekGbuz , I didn't put this check since it is not necessary here because every record has a 999 field with record id and instance id subfields. As well in this particular case to this method records that are passed always have related instances therefore the case when the 999 field has not instance id is not possible.

src/main/java/org/folio/service/export/LocalFileSystemExportService.java
```diff
@@ -1,62 +1,114 @@
...
+    JsonArray subfields = instanceIdHolderField.getJsonObject(GENERAL_INFO_FIELD_TAG_NUMBER).getJsonArray(SUBFIELDS_KEY);
+    return subfields.stream()
+      .map(JsonObject.class::cast)
+      .filter(jsonObject -> jsonObject.containsKey(INSTANCE_ID_KEY))
+      .findFirst().get().getString(INSTANCE_ID_KEY);
```
IlliaDaliek, line 109: The same here

alekGbuz (approved):

obozhko-folio (approved):

PR #254, opened 2021-07-30
MDEXP-425: When a highly inefficient query is done, vertx main thread is blocked 

IlliaDaliek:
Jira: https://issues.folio.org/browse/MDEXP-425

### PURPOSE
Make the code that requests instances for UUIDs to be executed in a worker pool instead of the main event loop thread.


alekGbuz (approved):

src/main/java/org/folio/service/file/upload/FileUploadServiceImpl.java
```diff
@@ -77,22 +78,35 @@
...
+      });
+    } else {
+      return updateFileDefinitionWithEmptyProgressIfAbsent(fileDefinition, params.getTenantId());
     }
-    if (Objects.isNull(fileDefinition.getJobExecutionId())) {
-      return updateFileDefinitionWithJobExecution(new JobExecution(), fileDefinition, params.getTenantId());
+  }
+
+  private Future<FileDefinition> updateFileDefinitionWithEmptyProgressIfAbsent(FileDefinition fileDefinition, String tenantId) {
+    if(Objects.isNull(fileDefinition.getJobExecutionId())) {
```
obozhko-folio, line 108: Could you please add space after 'if'?

obozhko-folio (approved):

src/main/java/org/folio/service/file/upload/FileUploadServiceImpl.java
IlliaDaliek: Sure

PR #258, opened 2021-08-11
MDEXP-425: When a highly inefficient query is done, vertx main thread is blocked 

IlliaDaliek:
Jira: https://issues.folio.org/browse/MDEXP-425

### PURPOSE

Avoid spamming of ThreadBlockedEception when the inventory request takes more than 60 seconds to be processed.  

### APPROACH

Refactor client to make it async and replace the verx.executeBlocking with a simple call to this new async method form.  
I tested the approach, it works fine.

### DIFF
--- a/src/main/java/org/folio/service/file/upload/FileUploadServiceImpl.java
+++ b/src/main/java/org/folio/service/file/upload/FileUploadServiceImpl.java
@@ -1,5 +1,6 @@
 package org.folio.service.file.upload;
 
+import io.vertx.config.ConfigRetriever;
 import io.vertx.core.Future;
 import io.vertx.core.Promise;
 import io.vertx.core.Vertx;
@@ -53,6 +54,8 @@ public class FileUploadServiceImpl implements FileUploadService {
   private ErrorLogService errorLogService;
   @Autowired
   private UsersClient usersClient;
+  @Autowired
+  private Vertx vertx;
 
   @Override
   public Future<FileDefinition> startUploading(String fileDefinitionId, String tenantId) {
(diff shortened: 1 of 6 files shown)


src/main/java/org/folio/service/file/upload/FileUploadServiceImpl.java
```diff
@@ -80,7 +85,9 @@
   public Future<FileDefinition> saveUUIDsByCQL(FileDefinition fileDefinition, String query, OkapiConnectionParams params) {
     if (StringUtils.isNotBlank(query)) {
       Promise<Optional<JsonObject>> instancesIdsPromise = Promise.promise();
-      Vertx.vertx().executeBlocking((blockingFeature) -> {
+      VertxOptions workerThreadOption = new VertxOptions().setMaxWorkerExecuteTimeUnit(TimeUnit.HOURS)
```
khandramai, line 88: workerThreadOption - need to be in plural for best readability

src/main/java/org/folio/service/file/upload/FileUploadServiceImpl.java
```diff
@@ -151,16 +147,18 @@
...
-      saveUUIDsFromJson(instancesUUIDs.get(), fileDefinition, jobExecution, request, params)
-        .onSuccess(promise::complete)
-        .onFailure(ar -> failFileDefinitionAndJobExecution(promise, fileDefinition, jobExecution, request, ar.getCause(), params));
-    } else {
-      saveUUIDsFromJson(new JsonObject(), fileDefinition, jobExecution, request, params)
-        .onSuccess(promise::complete)
-        .onFailure(ar -> failFileDefinitionAndJobExecution(promise, fileDefinition, jobExecution, request, ar.getCause(), params));
-    }
+    inventoryClient.getInstancesBulkUUIDsAsync(request.getCriteria(), params).onComplete(instancesUUIDsResult -> {
+      Optional<JsonObject> instancesUUIDs = instancesUUIDsResult.result();
```
IlliaDaliek, line 151: Here the Optional object always will be returned, no need into onSuccess/onFailure methods

alekGbuz (approved):

obozhko-folio (approved):

PR #289, opened 2021-11-10
MDEXP-473 - CQL export fails when effective location is included in the cql statement

siarhei-charniak:
[MDEXP-473](https://issues.folio.org/browse/MDEXP-473) - CQL export fails when effective location is included in the cql statement

## Purpose
The export fails when triggered by CQL and the effective location is a part of the statement. If works fine if the location is not present

## Approach
* Introduced mod-search client to leverage /search/instances/ids API instead of mod-inventory storage API to get instance IDs
* Improved error log
* Updated unit tests

## Pre-Merge Checklist

### DIFF
--- a/src/main/java/org/folio/clients/SearchClient.java
+++ b/src/main/java/org/folio/clients/SearchClient.java
@@ -1,0 +1,99 @@
+package org.folio.clients;
+
+import static java.lang.String.format;
+import static org.folio.rest.RestVerticle.OKAPI_HEADER_TENANT;
+import static org.folio.rest.RestVerticle.OKAPI_HEADER_TOKEN;
+import static org.folio.util.ExternalPathResolver.SEARCH_IDS;
+import static org.folio.util.ExternalPathResolver.resourcesPathWithPrefix;
+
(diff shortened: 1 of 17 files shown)


src/main/java/org/folio/clients/SearchClient.java
```diff
@@ -75,8 +78,13 @@ public SearchClient(WebClient webClient, ErrorLogService errorLogService) {
         } else {
           try {
             JsonObject instances = response.bodyAsJsonObject();
-            if (instances != null) {
-              promise.complete(Optional.of(instances));
+            if (nonNull(instances)) {
+              if (instances.getJsonArray("ids").isEmpty()) {
```
khandramai, line 82: Is it make sense to save "ids" as string constant?

khandramai (approved):

src/main/java/org/folio/clients/SearchClient.java
siarhei-charniak: Good catch, fixed

obozhko-folio (approved):

PR #298, opened 2021-12-17
MDEXP-480: Only one mapping is executed during the export

IlliaDaliek:
Jira: https://issues.folio.org/browse/MDEXP-480

### PURPOSE

Recognize fields as different when they have the same field id but different indicators.
Adjust the tests to the corrected business logic.

### APPROACH

When building a mapping rule take the indicators into account instead of relying on the field tag only.


pom.xml
```diff
@@ -27,6 +27,7 @@
     <ramlfiles_path>${basedir}/ramls</ramlfiles_path>
     <postgres-testing.version>33.0.0</postgres-testing.version>
     <minio.version>8.3.0</minio.version>
+    <generate-marc-utils-version>1.4.0-SNAPSHOT</generate-marc-utils-version>
```
IlliaDaliek, line 30: Since the library with required changes wasn't released yet we have to specify the SNAPSHOT version to have latest library changes available

alekGbuz (approved):

siarhei-charniak (approved):

PR #300, opened 2022-01-17
MDEXP-303: Generate MFHD records on the fly

IlliaDaliek:
Jira: https://issues.folio.org/browse/MDEXP-303

### PURPOSE

Generate MFHD records on the fly using default holdings mapping rules in case of an absent underlying srs record.

### APPROACH

Since the business logic for mapping JSON using rules already presented, in sope of this story, I adapted holdings export to use the same code. 
So, the flow in case of absent srs record for holding is next:
1) we request holdings by id
2) request instance for each holding (here the instance is required for appending instance hrid to a generated srs record)
3) pass the JSON to an existent mapping related code


src/main/java/org/folio/service/mapping/converter/RecordConverter.java
```diff
@@ -1,62 +1,94 @@
...
+    return holdings.stream()
+      .map(JsonObject.class::cast)
+      .map(json -> json.getString("instanceId"))
+      .collect(Collectors.toList());
+  }
+
+  /**
+   * Fetches items if Transformations specify Record type "ITEM"
+   */
+  protected void fetchItemsIfRequired(MappingProfile mappingProfile, OkapiConnectionParams params, String holdingUUID, JsonObject appendHoldingsItems, String jobExecutionId) {
```
IlliaDaliek, line 83: Items aren't needed for now but I left the code since it will be required when supporting custom mp will be implemented.

alekGbuz (approved):

obozhko-folio (approved):

src/main/java/org/folio/service/loader/RecordLoaderServiceImpl.java
```diff
@@ -33,46 +39,62 @@ public RecordLoaderServiceImpl(@Autowired SourceRecordStorageClient srsClient, @
...
+  private void populateLoadResultFromInventory(Collection<String> entityIds, JsonObject entities, LoadResult loadResult) {
     List<JsonObject> inventoryRecords = new ArrayList<>();
-    Set<String> singleInstanceIdentifiersSet = new HashSet<>(instanceIds);
-    for (String instanceId : instanceIds) {
-      for (Object instance : instanceRecord.getJsonArray("instances")) {
-        JsonObject record = (JsonObject) instance;
-        if (record.getValue("id").equals(instanceId)) {
-          singleInstanceIdentifiersSet.remove(instanceId);
+    Set<String> entitiesIdentifiersSet = new HashSet<>(entityIds);
+    String jsonArrayKey = loadResult.getEntityType().equals(AbstractExportStrategy.EntityType.INSTANCE) ? INSTANCES : HOLDINGS_RECORDS;
```
alekGbuz, line 83: From my point of view, it is better to use == for enum comparing.

PR #303, opened 2022-01-20
MDEXP-434: Generate MARC bib record - Identifiers  part VI 

IlliaDaliek:
Jira: https://issues.folio.org/browse/MDEXP-434

### PURPOSE

FOLIO inventory storage instance contains fields that are represented by keys and those will need to be replaced by their values when generating MARC bib record on the fly. In many cases, which field in MARC record will be populated depends on the value of the key. For identifiers, depending on its type the different MARC Bib record field will be populated. Each inventory instance has a list of identifiers associated with it and it is an array of identifier value and identifier type id.

### APPROACH
Updated transformation rules
Updated unit test
Verified locally (using the Vagrant box):

**Prepared an instance with identifiers**

inst

**The exported record**

record result


IlliaDaliek:
Will push a unit tests soon

siarhei-charniak (approved):

